PR #7, opened 2019-10-02
Validate only the input manifest

howardjohn:
The standard manifest is not a user input, so no need to validate it.
This is triggered by the fact that the validation is not actually
correct for the standard manifest, as it has no requirement for the git
source specified.


howardjohn:
Builds are blocked: https://prow.istio.io/view/gcs/istio-prow/logs/release_istio_postsubmit/4

howardjohn:
Added a new test to actually run the release, which passes now with this commit

PR #25, opened 2019-10-18
Use service account auth instead of docker

howardjohn:
https://prow.istio.io/view/gcs/istio-prow/logs/build-release_release-builder_postsubmit/1 failed due to bad credential set up. This is what the old release did, so it should work...


howardjohn:
Note: I also added a newline to trigger the build again

PR #49, opened 2019-10-29
fix test-infra branch and retrigger build

ericvn:


geeknoid:
We actually should have branched gogo-genproto. Should we do that now?

ericvn:
@geeknoid `gogo-genproto` branched and changed here.

PR #54, opened 2019-11-04
Add operator deploy/ and samples/

howardjohn:
For https://github.com/istio/istio/issues/18441

Does not fully fix since istio.io needs updates

@sdake

### DIFF
--- a/pkg/build/archive.go
+++ b/pkg/build/archive.go
@@ -67,11 +67,12 @@ func Archive(manifest model.Manifest) error {
 		if err := util.CopyDirFiltered(path.Join(manifest.RepoDir("istio"), "install"), path.Join(out, "install"), includePatterns); err != nil {
 			return err
 		}
+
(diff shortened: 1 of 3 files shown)


pkg/build/archive.go
```diff
@@ -80,6 +81,15 @@ func Archive(manifest model.Manifest) error {
...
 		}
+		if err := util.CopyDir(path.Join(manifest.RepoDir("operator"), "deploy"), path.Join(out, "install/kubernetes/operator/deploy")); err != nil {
+			return err
+		}
+		if err := sanitizeTemplate(manifest, path.Join(out, "install/kubernetes/operator/deploy/operator.yaml")); err != nil {
+			return fmt.Errorf("failed to sanitize operator manifest")
+		}
+		if err := util.CopyDir(path.Join(manifest.RepoDir("operator"), "samples"), path.Join(out, "samples/operator")); err != nil {
+			return err
+		}
```
jwendell: Does it make sense to sanitize the samples folder as well? It currently doesn't have any hub or tag but it could be added in the future, no? Then we would be safe.

PR #64, opened 2019-11-11
Use BUILDINFO file on build

jwendell:
This way we can override the variables that appear on
`version` subcommand of binaries, avoiding having
`-dirty` or `Modified` to be in the output.

Closes https://github.com/istio/istio/issues/18636

### DIFF
--- a/pkg/source.go
+++ b/pkg/source.go
@@ -25,6 +25,8 @@ import (
 	"istio.io/release-builder/pkg/model"
 	"istio.io/release-builder/pkg/util"
 
+	"io/ioutil"
+
 	"istio.io/pkg/log"
 )
 
@@ -133,5 +135,36 @@ func StandardizeManifest(manifest *model.Manifest) error {
 		}
 		manifest.Dependencies.Set(repo, newDep)
 	}
+	return nil
+}
+
+// GenerateBuildInfoFile creates the BUILDINFO file, used by build scripts
+func GenerateBuildInfoFile(manifest *model.Manifest) error {
+	content := fmt.Sprintf(
+		`istio.io/pkg/version.buildVersion=%s
+istio.io/pkg/version.buildGitRevision=%s
+istio.io/pkg/version.buildStatus=Clean
+istio.io/pkg/version.buildTag=%s
+istio.io/pkg/version.buildHub=%s
+`,
+		manifest.Version,
+		manifest.Dependencies.Istio.Sha,
+		manifest.Version,
+		manifest.Docker)
+
+	f, err := ioutil.TempFile("", "BUILDINFO_")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.WriteString(content)
+	if err != nil {
+		return err
+	}
+
+	manifest.BuildInfoFileName = f.Name()
+	log.Infof("Created buildInfo file: %s", manifest.BuildInfoFileName)
+
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/source.go
```diff
@@ -135,3 +137,34 @@ func StandardizeManifest(manifest *model.Manifest) error {
...
+func GenerateBuildInfoFile(manifest *model.Manifest) error {
+	content := fmt.Sprintf(
+		`istio.io/pkg/version.buildVersion=%s
+istio.io/pkg/version.buildGitRevision=%s
+istio.io/pkg/version.buildStatus=Clean
+istio.io/pkg/version.buildTag=%s
+istio.io/pkg/version.buildHub=%s
+`,
+		manifest.Version,
+		manifest.Dependencies.Istio.Sha,
```
howardjohn: This will give us the wrong version for dependencies from operator and CNI I think

jwendell: Well, currently only Istio build system makes use of BUILDINFO.

cni does not even have a version flag.

operator currently outputs something like `version.BuildInfo{Version:"unknown", GitRevision:"unknown", GolangVersion:"go1.13.4", BuildStatus:"unknown", GitTag:"unknown"}`

howardjohn: > operator currently outputs something like version.BuildInfo{Version:"unknown", GitRevision:"unknown", GolangVersion:"go1.13.4", BuildStatus:"unknown", GitTag:"unknown"}

:slightly_frowning_face: they should be using it

jwendell: yeah. I can drop this and just remove the dirty/Modified checks in istio/istio.

I kind of like them though. I think they can be useful when people are building on their own (not useful to our official releases though).

pkg/source.go
```diff
@@ -135,3 +137,34 @@ func StandardizeManifest(manifest *model.Manifest) error {
...
+		return err
+	}
+	defer f.Close()
+
+	_, err = f.WriteString(content)
+	if err != nil {
+		return err
+	}
+
+	manifest.BuildInfoFileName = f.Name()
```
howardjohn: I think .Name() is not the full path so this might not work?

pkg/source.go
```diff
@@ -135,3 +137,34 @@ func StandardizeManifest(manifest *model.Manifest) error {
...
+istio.io/pkg/version.buildStatus=Clean
+istio.io/pkg/version.buildTag=%s
+istio.io/pkg/version.buildHub=%s
+`,
+		manifest.Version,
+		manifest.Dependencies.Istio.Sha,
+		manifest.Version,
+		manifest.Docker)
+
+	f, err := ioutil.TempFile("", "BUILDINFO_")
```
howardjohn: Can we write this somewhere in `manifest.WorkDir()` instead of a new tmp dir?

pkg/source.go
howardjohn: ok, we should make sure we update this to pass in a SHA if operator and/or CNI fix this. I will report issues for both

howardjohn: CNI doesn't have a version. Operator recently fix this

howardjohn:
I am fine with any of 
* This change
* Drop the dirty flag for everyone
* Env var to skip dirty flag

jwendell:
I liked the option **Env var to skip dirty flag**. It's way cleaner than this one.
Opened https://github.com/istio/common-files/pull/290 and https://github.com/istio/release-builder/pull/349.

PR #73, opened 2019-11-15
Override proxy by url rather than local file

howardjohn:
This supports private builds easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ dependencies:
   proxy:
     git: https://github.com/istio/proxy
     auto: deps
-# proxyOverride specifies an Envoy binary to use, instead of pulling one from GCS
-proxyOverride: /path/to/envoy
+# proxyOverride specifies an alternative URL to pull Envoy binary from
+proxyOverride: https://storage.googleapis.com/istio-build/proxy
 ```
 
 ## Publish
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -58,8 +58,8 @@ dependencies:
   proxy:
     git: https://github.com/istio/proxy
     auto: deps
-# proxyOverride specifies an Envoy binary to use, instead of pulling one from GCS
-proxyOverride: /path/to/envoy
+# proxyOverride specifies an alternative URL to pull Envoy binary from
+proxyOverride: https://storage.googleapis.com/istio-build/proxy
```
kyessenov: Do you think we should plan to replace this workflow with docker images in 1.5?
It's dangerous to use x86 binary in an arbitrary docker context. Differences in gblic / libc++ can be quite subtle. I think validating the working proxy docker image should be left to envoy/proxy, not here.

howardjohn: Seems reasonable to me, but I am not familiar with all the details about where it could go wrong. I think a lot of our tests directly run the Envoy binary so it would break all of that? If we can get it working sgtm

We do still need this change though, for building 1.4

kyessenov: The only tests that do that are 1) mixer tests (which we'll deprecate eventually) 2) envoy agent tests. That's not that many. Agent tests could be moved to proxy repo.

howardjohn: I think pilot are doing this as well

PR #88, opened 2019-12-02
Build Istio 1.4.1

ericvn:


howardjohn:
Retesting just to sanity check the right SHA will be built; the warning outputs what sha it builds

PR #90, opened 2019-12-06
Prevent nil pointer in TestManifest

clarketm:
When only the required dependencies are supplied in the **manifest** (i.e. istio, cni, operator, proxy) the *validation* fails.

> This may not be the correct approach. Alternatively, a flag (i.e. `--minimal`) or just including *all* the dependencies may be better.

> Also, I have yet to verify if other tests need an update as well, but I surmise at least some others, like `TestLicenses`, will fail as well.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -283,7 +283,10 @@ func TestOperator(r ReleaseInfo) error {
 func TestManifest(r ReleaseInfo) error {
 	for _, repo := range []string{"api", "cni", "client-go", "istio", "operator", "pkg", "proxy"} {
 		d, f := r.manifest.Dependencies.Get()[repo]
-		if !f || d.Sha == "" {
+		if !f || d == nil {
+			continue
+		}
+		if d.Sha == "" {
 			return fmt.Errorf("got empty SHA for %v", repo)
 		}
 	}


howardjohn:
@clarketm I think my original suggestion to just include these 3 may not have been the best after I think about it some more.

Whats happening today is we build with our 3 private repos (good) and produce a release. This release does not have the "extras".

Then we copy this to public, publish it in public. The publish does NOT tag the extra repos (so we would need to manually), and does not include the licenses (potential legal issues?).


So I think we have three options
1. Add all the repos, pointing to private. This isn't too bad but we have to sync all the repos which is painful.
2. Add the 3 core repos pointing to private, and the extras pointing to public. 
3. We build in private only for testing. Then we upstream the patches, build in public, and publish in public. This would add 1hr delay to the release if we got everything right, and if we had issues with syncing the repos it could be much longer


Based on these I think 2 might make the most sense? What do you think?

clarketm:
> Based on these I think 2 might make the most sense? What do you think?

I agree. Number **2** has the same benefits as number **1** (i.e. a complete *private* release) without the need to concern ourselves with synching repos or convoluting the existing release validation. 

I can update the manifest secret with the other repos and rerun the release job. I should pull these values directly from `istio/release-builder@release-1.4` upstream correct? 

**e.g.**
https://github.com/istio/release-builder/blob/a73cf7a9cbe265986e0750b7fbac8e4dc009ea4e/release/build.sh#L60-L83

howardjohn:
One problem will be

```
  proxy:
    git: https://github.com/istio/proxy
    auto: deps
```

The `auto: deps` is an obscure way of saying "read istio.deps from istio/istio, and use that SHA". That SHA won't exist in public, so that one will actually need to be private

Otherwise LGTM

clarketm:
> One problem will be
> 
> ```
>   proxy:
>     git: https://github.com/istio/proxy
>     auto: deps
> ```
> 
> The `auto: deps` is an obscure way of saying "read istio.deps from istio/istio, and use that SHA". That SHA won't exist in public, so that one will actually need to be private

For `proxy`, we are using the `PROXY_OVERRIDE`, so this entry can continue to be safely omitted?

howardjohn:
But we still want to get the licenses and tag the proxy repo. Adding it to this list is doing only that, it won't actually be used as part of the build. I realize this is confusing :slightly_smiling_face:

clarketm:
> But we still want to get the licenses and tag the proxy repo. Adding it to this list is doing only that, it won't actually be used as part of the build. I realize this is confusing 🙂

Make sense now, thanks.

PR #91, opened 2019-12-06
Hide proxy override from output

howardjohn:
No need for this to be in the final output


clarketm:
Should this be cherry-picked to `release-1.4` as well or is that branch frozen?

PR #100, opened 2019-12-16
Tag repo with Go compatible version names

jasonwzm:
fix https://github.com/istio/api/issues/1201

The fix does not do double tagging. If a repo needs to be exposed for external use, we will just release it with go version convention.


pkg/manifest.go
```diff
@@ -107,5 +108,8 @@ func ReadInManifest(manifestFile string) (model.InputManifest, error) {
 	if err := validateManifestDependencies(manifest.Dependencies); err != nil {
 		return manifest, fmt.Errorf("invalid manifest: %v", err)
 	}
+	if _, err = semver.StrictNewVersion(manifest.Version); err != nil {
```
howardjohn: I think this will fail for 1.5-dev.sha

jasonwzm: can we name the release 1.5.0-dev.sha? So that all of our versions in the future will be semver.

howardjohn:
it's not accurate then. Consider the 1.4 cases. If we named a PR at head of
1.4 to be 1.4.0 then it's wrong, it's after 1.4.2. But we can't just name
it 1.4.3 since we don't know patch release info.

Simply adding patch release info to the repo won't work either since commit
order is not necessarily release order (confusingly) when we have security
patches in play, which is often. I can point you to a slack thread with a
lot of info if you are interested.

If there is a good solution though that would be great

jasonwzm:
@howardjohn If you can point me to those conversations, that'd be great! But from my understanding, with our current branching strategy, any new PRs that are merged will belong to current patch version + 1. For example, any new PR merged now will be built with version 1.4.3-dev.sha?

jasonwzm:
@howardjohn Thanks! Based on the scenarios discussed in the slack chat, I removed the semver check for the input version and moved it to where we tag go module version.

PR #103, opened 2019-12-18
Update common files

clarketm:


howardjohn:
this is using the old image in prow

clarketm:
> this is using the old image in prow

https://github.com/istio/test-infra/pull/2235

PR #105, opened 2019-12-19
Expect debian binaries in the TARGET_OUT directory

sdake:
This PR enables the merging of https://github.com/istio/istio/pull/19322.

This PR causes the debian package to be copied from two locations, as is done in the docker.go implementation.

### DIFF
--- a/pkg/build/debian.go
+++ b/pkg/build/debian.go
@@ -27,10 +27,8 @@ func Debian(manifest model.Manifest) error {
 	if err := util.RunMake(manifest, "istio", nil, "sidecar.deb"); err != nil {
 		return fmt.Errorf("failed to build sidecar.deb: %v", err)
 	}
-	if util.FileExists(path.Join(manifest.RepoOutDir("istio"), "deb")) {
-		if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
-			return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
-		}
+	if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
+		return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/build/debian.go
```diff
@@ -27,11 +27,16 @@ func Debian(manifest model.Manifest) error {
...
+		if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
+			return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
+		}
 	}
 	if err := util.CreateSha(path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
 		return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
 	}
+	if err := util.CopyFilesToDir(path.Join(manifest.GoOutDir(), "deb"), path.Join(manifest.OutDir(), "deb")); err != nil {
+		return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
+	}
```
howardjohn: why do we have this? isn't the above change sufficient?

sdake: 8am? :)

sdake: I put this  CopyFilesToDir in the wrong place. This is simply needed to make the dryrun check pass and not block the gate. We can merge without it if you like. Please advise.

pkg/build/debian.go
```diff
@@ -27,8 +27,10 @@ func Debian(manifest model.Manifest) error {
 	if err := util.RunMake(manifest, "istio", nil, "sidecar.deb"); err != nil {
 		return fmt.Errorf("failed to build sidecar.deb: %v", err)
 	}
-	if err := util.CopyFile(path.Join(manifest.GoOutDir(), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
-		return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
+	if util.FileExists(path.Join(manifest.RepoOutDir("istio"), "deb")) {
```
sdake: this conditional is also not correct. If you want me to change the PR to read from both locations so the checks all pass, I can do that. If you want me to read from one place (as htis pr does with the absence of htis conditional) - the dryrun check will fail.

howardjohn: let's just keep it simple and I'll merge this?

sdake: i already pushed a change..  but put a hold on it ;)

PR #110, opened 2019-12-23
fix not set GOPATH

helight:
if there is not set GOPATH
publish.sh will report error:
./test/publish.sh: line 77: GOPATH: unbound variable

from go 1.12, there is no need to set GOPATH for go.

### DIFF
--- a/test/publish.sh
+++ b/test/publish.sh
@@ -74,7 +74,9 @@ EOF
 )
 
 # "Temporary" hacks
-export PATH=${GOPATH}/bin:${PATH}
+if [[ -z ${GOPATH} ]]; then
+  export PATH=${GOPATH}/bin:${PATH}
+fi
 
 go run main.go build --manifest <(echo "${MANIFEST}")
 


test/publish.sh
```diff
@@ -74,7 +74,9 @@ EOF
 )
 
 # "Temporary" hacks
-export PATH=${GOPATH}/bin:${PATH}
+if [[ -z ${GOPATH} ]]; then
```
howardjohn: because of `set -eux` this will still fail if gopath is not set


I highly recommend you run this in a docker container so you build with the same environment as we do in official releases

helight: OK. I did it run it in a docker container yesterday,  or did you mean I should run this with gcr.io/istio-testing/build-tools?

howardjohn:
Yes inside build-tools if you want to replicate

helight:
> Yes inside build-tools if you want to replicate
OK, I will try it today, thanks.

PR #116, opened 2020-01-21
Changes required to do a release build on a Mac

ericvn:
Changes I needed to do a release build on my Mac.

I'm not sure necessarily why some of the `make` output paths seem to be different, but this is what I needed to do to get past the `go run main.go build --manifest <manifest>`. 

I still need to run the _validate_

### DIFF
--- a/pkg/build/archive.go
+++ b/pkg/build/archive.go
@@ -103,7 +103,7 @@ func Archive(manifest model.Manifest) error {
 			istioctlBinary += ".exe"
 			istioctlDest += ".exe"
 		}
-		if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), istioctlBinary), path.Join(out, "bin", istioctlDest)); err != nil {
+		if err := util.CopyFile(path.Join(os.Getenv("TARGET_OUT"), "release", istioctlBinary), path.Join(out, "bin", istioctlDest)); err != nil {
 			return err
 		}
 		if err := os.Chmod(path.Join(out, "bin", istioctlDest), 0755); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/build/archive.go
```diff
@@ -103,7 +103,7 @@ func Archive(manifest model.Manifest) error {
 			istioctlBinary += ".exe"
 			istioctlDest += ".exe"
 		}
-		if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), istioctlBinary), path.Join(out, "bin", istioctlDest)); err != nil {
+		if err := util.CopyFile(path.Join(os.Getenv("TARGET_OUT"), "release", istioctlBinary), path.Join(out, "bin", istioctlDest)); err != nil {
```
howardjohn: this does not seem right, but maybe I am understanding wrong. This environment variable is only set in istio repo not here.

And as should always be copying Linux stuff here

ericvn: TARGET_OUT is set in the docker run command in the common Makefile. Maybe it's a misunderstanding by me on how the building container should be `docker run` (I'm basically doing what the common Makefile does).

Right now I'm chasing an issue in the validate where it seems there is a missing _proxyv2.tar.gz_, so could certainly be something wrong with changes as they are now.

ericvn:
Looking at the _dry-run_ test I can see that I am indeed seeing something different in the build between my Mac and what prow sees (ie. things are built in different directories so my changes in this PR to pull from different locations needed for my Mac is different than the infrastructure. Example, the _dry-run failure has
```
2020-01-21T04:53:19.532450Z	info	Copying release/istio-sidecar.deb -> /tmp/tmp.EsmTP5dcw1/build/out/deb/istio-sidecar.deb
Error: failed to build: failed to build Debian: failed to package istio-sidecar.deb: failed to open file release/istio-sidecar.deb to copy: open release/istio-sidecar.deb: no such file or directory
```
which was a location I changed in this PR since the file was created in a different location locally.

ericvn:
Found I can get this to create a release if I don't set all the env vars I was setting in the container (ie, remove the various TARGET ones). As such I'll close this PR.

PR #153, opened 2020-02-27
Add grafana to build step

howardjohn:
This currently isn't wired to push the dashboards yet, because we
haven't put the credentials in the cluster. It does everything but that
part though, and validates things are set up correctly. I have tested
this with my own grafana account. For
istio/istio#21541

### DIFF
--- a/pkg/build/grafana.go
+++ b/pkg/build/grafana.go
@@ -1,0 +1,136 @@
+// Copyright Istio Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


pkg/build/grafana.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	msg["__input"] = inputMsg
+
+	var requiresMesh json.RawMessage
+	if err := json.Unmarshal([]byte(`[
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "6.4.3"
```
douglas-reid: should this be another param ?

howardjohn: I basically took our dashboards, clicked export, and looked at what they added. All of them seemed to have this version, so I assume we use some panel or feature in that version? The only thing we will need to do is ensure that if we utilize a new feature we update this. I think the impact would just be that someone with an old grafana deployment may unexpectedly have some panel break though?

Any other ideas to improve this are welcome, this is a huge hack. I don't really want to try to spin up a whole grafana instance and extract this each build as well, seems a bit heavy-handed

PR #207, opened 2020-04-22
Trigger 1.6 alpha 2 build

brian-avery:


jacob-delgado:
Fixes https://github.com/istio/istio/issues/22814

PR #224, opened 2020-05-04
Include samples/certs/Makefile in distribution

nmittler:
This is needed by the new docs for setting up trust across clusters.


myidpt:
Thanks for the work!

PR #242, opened 2020-05-13
Generate compiled in charts for operator binary

ostromart:
This runs make gen-charts before the docker targets to generate the correct compiled-in charts for the operator binary.

### DIFF
--- a/pkg/build/docker.go
+++ b/pkg/build/docker.go
@@ -34,6 +34,12 @@ func Docker(manifest model.Manifest) error {
 	}
 
 	for _, repo := range []string{"istio", "cni"} {
+		if repo == "istio" {
+			// Istio operator requires compiled-in charts to be generated before the image is built.
+			if err := util.RunMake(manifest, repo, env, "gen-charts"); err != nil {
+				return fmt.Errorf("failed to make istio gen-charts: %v", err)
+			}
+		}
 		if err := util.RunMake(manifest, repo, env, "docker.save"); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/build/docker.go
```diff
@@ -34,6 +34,12 @@ func Docker(manifest model.Manifest) error {
 	}
 
 	for _, repo := range []string{"istio", "cni"} {
+		if repo == "istio" {
+			// Istio operator requires compiled-in charts to be generated before the image is built.
+			if err := util.RunMake(manifest, repo, env, "gen-charts"); err != nil {
```
howardjohn, line 39: Is there anyway to make this a step done as part of `docker.operator`? That seems like a more logical place to add it. If you ever run `docker.operator` without this its broken, right? If not let me know

We may always want the docker image to use charts baked into the dockerfile instead of compiled in

ostromart: my concern is that docker.operator is run as part of top level docker which will happen often. so devs that have nothing to do with operator will start seeing assets.gen.go appearing.

ostromart: i'd rather inconvenience a few operator devs that will get used to the extra step quickly

howardjohn: Ok I was thinking that a bit as well, makes sense. lets do this.

How does the operator test pass though?

ostromart: that's a great question for @richardwxn :-)

howardjohn:
I think th test may call gen

PR #263, opened 2020-06-03
Add /tools/certs/ to the release package.

myidpt:


myidpt:
@howardjohn Thank you for the comment.

PR #271, opened 2020-06-09
Scan base image for vulerabilities during build

ericvn:
Part of work related to https://github.com/istio/istio/issues/23243.


ericvn:
One comment from the PS WG yesterday was the addition of a switch to turn off the failing of the build for a base image vulnerability to be used in the case where there is an emergency build needed and no time to get a new istio/base built and Makefile updated to use it.

PR #281, opened 2020-06-16
CNI is no longer a dependency, it's part of istio/istio

jwendell:
Part of https://github.com/istio/istio/issues/24610

### DIFF
--- a/pkg/build/docker.go
+++ b/pkg/build/docker.go
@@ -47,9 +47,5 @@ func Docker(manifest model.Manifest) error {
 		}
 	}
 
-	// Others output to GoOut, so copy those as well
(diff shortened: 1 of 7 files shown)


pkg/build/docker.go
```diff
@@ -33,23 +33,20 @@ func Docker(manifest model.Manifest) error {
...
+		return fmt.Errorf("failed to create %v docker archives: %v", "istio", err)
+	}
+	if util.FileExists(path.Join(manifest.RepoOutDir("istio"), "docker")) {
+		// Some repos output docker files to the source repo
+		if err := util.CopyFilesToDir(path.Join(manifest.RepoOutDir("istio"), "docker"), path.Join(manifest.OutDir(), "docker")); err != nil {
+			return fmt.Errorf("failed to package docker images: %v", err)
 		}
 	}
+
 	// Others output to GoOut, so copy those as well
```
howardjohn, line 50: this is probably dead now actually, but not related to this PR

PR #313, opened 2020-07-10
Add RPMs to the build

howardjohn:
Also fixed a bug in the debian one (it called a non-existant target, but
it happened to pass because we already call it in the docker step), and
added tests

For https://github.com/istio/istio/issues/9117

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -54,6 +54,12 @@ func Build(manifest model.Manifest) error {
 		}
 	}
 
+	if _, f := manifest.BuildOutputs[model.Rpm]; f {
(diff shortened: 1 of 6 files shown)


brian-avery:
@howardjohn 

```
2020-07-10T21:48:55.823959Z	info	Check failed: check Rpm failed: rpm package not found
Error: release validation FAILED
exit status 1
++ docker ps -q
+ docker kill
"docker kill" requires at least 1 argument.
See 'docker kill --help'.
```

Is this related?

howardjohn:
That is a test working as intended :slightly_smiling_face: forgot to actually call the function

PR #320, opened 2020-07-15
Remove operator deploy verification

sdake:


sdake:
https://github.com/istio/istio/issues/27139

PR #335, opened 2020-07-27
Add envoy into deps and proxy_workspace auto resolving

bianpengyuan:
ref: https://github.com/istio/istio/issues/25866

This is to make release builder tag istio/envoy repo as other repos.

proxy PR to add presubmit check on presence of `ENVOY_SHA` in workspace file: https://github.com/istio/proxy/pull/2948

### DIFF
--- a/pkg/util/files.go
+++ b/pkg/util/files.go
@@ -16,7 +16,6 @@ package util
 
 import (
 	"archive/zip"
-	"bufio"
 	"crypto/sha256"
 	"encoding/json"
 	"errors"
@@ -27,6 +26,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/rogpeppe/go-internal/modfile"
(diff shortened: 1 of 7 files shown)


pkg/util/files.go
```diff
@@ -222,6 +226,43 @@ func fetchAutoDeps(repo string, dep *model.Dependency, dest string) error {
 	return nil
 }
 
+func fetchAutoProxyWorkspace(dep *model.Dependency, dest string) error {
+	wsFile, err := os.Open(path.Join(dest, "../proxy/WORKSPACE"))
```
jwendell, line 230: I think the code below could be simplified by using `ioutil.ReadFile` and [regular expressions](https://golang.org/pkg/regexp/#Regexp.FindString).

test/publish.sh
```diff
@@ -65,6 +65,9 @@ dependencies:
   tools:
     git: https://github.com/istio/tools
     branch: master
+  envoy:
+    git: https://github.com/envoyproxy/envoy-wasm
```
jwendell, line 69: Shouldn't this be `istio/envoy`?

jwendell:
We currently don't maintain `istio/envoy` master in sync with upstream. Until a branch is cut, the SHA in WORKSPACE points directly to upstream. That's the reason the test failed. It should work fine on release-1.7 though.

We do not often create releases based on master, so, this isn't a big deal IMO. If such a release is going to be made, we can just sync `istio/envoy` master with upstream and this test is going to pass.

pkg/util/files.go
bianpengyuan: Thanks, done.

test/publish.sh
bianpengyuan: This is because of the same reason you mentioned here: https://github.com/istio/release-builder/pull/335#issuecomment-664351623. Dry run will fail if this is set to istio/envoy. I think it is fine to use upstream envoy repo here, continuously syncing istio/envoy master just to make dry-run test pass seems unnecessary.

jwendell: ah ok, this is used only in testing, right? We should make sure to not try to push any tag to it :)

bianpengyuan:
@jwendell Thanks! I will do a manual cherry-picking.

PR #348, opened 2020-07-31
Fix dirty state

jwendell:
This way we avoid running `make gen-charts` only at release build script,
thus making the working tree dirty. This causes commands like `<binary> version`
to report a dirty, modified state.

This change makes sure the tree is clean before building the binaries.

### DIFF
--- a/pkg/build/docker.go
+++ b/pkg/build/docker.go
@@ -33,9 +33,9 @@ func Docker(manifest model.Manifest) error {
 		env = append(env, "ISTIO_ENVOY_BASE_URL="+manifest.ProxyOverride)
 	}
 
-	// Istio operator requires compiled-in charts to be generated before the image is built.
-	if err := util.RunMake(manifest, "istio", env, "gen-charts"); err != nil {
-		return fmt.Errorf("failed to make istio gen-charts: %v", err)
+	// Make sure operator's compiled-in charts are up to date before the image is built.
+	if err := util.RunMake(manifest, "istio", env, "gen-charts", "check-clean-repo"); err != nil {
+		return fmt.Errorf("failed to verify charts are up to date. Run `make gen-charts` in istio/istio and commit the results before making a release: %v", err)
 	}
 	if err := util.RunMake(manifest, "istio", env, "docker.save"); err != nil {
 		return fmt.Errorf("failed to create %v docker archives: %v", "istio", err)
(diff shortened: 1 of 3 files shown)


pkg/build/docker.go
```diff
@@ -33,9 +33,9 @@ func Docker(manifest model.Manifest) error {
 		env = append(env, "ISTIO_ENVOY_BASE_URL="+manifest.ProxyOverride)
 	}
 
-	// Istio operator requires compiled-in charts to be generated before the image is built.
-	if err := util.RunMake(manifest, "istio", env, "gen-charts"); err != nil {
-		return fmt.Errorf("failed to make istio gen-charts: %v", err)
+	// Make sure operator's compiled-in charts are up to date before the image is built.
+	if err := util.RunMake(manifest, "istio", env, "gen-charts", "check-clean-repo"); err != nil {
```
howardjohn, line 37: won't check-clean-repo fail after `gen-charts`?

jwendell: that's the idea :)

jwendell: And this failure is legit, until we fix it in istio/istio (in 1.7 branch)

jwendell:
@howardjohn we can't make `dry-run_release-builder_release-1.7` required unless it runs for *every* PR (instead of based on a regex of changed files), right?

howardjohn:
@jwendell yes, it just says its not Required by github though; prow internally treats it as required though

jwendell:
Ah ok. Cool. IMO we could run it for every PR, but setting `DRY_RUN` env variable to 1 to avoid the publishing.

jwendell:
Copied the commit message into the PR description. This closes https://github.com/istio/istio/issues/18636.
@howardjohn thoughts?

jwendell:
This means that before a release is made, operator charts need to be updated, which makes sense, right?

howardjohn:
There is a reason we don't commit the results of the charts, it leads to major merge conflicts and developer pain. I don't like it either but IMO the right solution is to get rid of the codegen entirely which we agreed to do in environments WG but started getting push back on the PRs

jwendell:
@howardjohn so, is this risky, dangerous even on a branch (i.e., not on master)?

howardjohn:
I think it will break any cherrypick changing the install is one concern

jwendell:
ok, I think I'll resurrect https://github.com/istio/release-builder/pull/64 then

PR #357, opened 2020-08-03
Fix client-go versioning

howardjohn:
Previously we intended to change the version from 1.7.0 to v1.7.0, but
we didn't actually do this.

On hold for @jasonwzm to verify we still desire this. Note - we never successfully tagged with `v` prefix

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -100,7 +100,7 @@ func (i IstioDependencies) MarshalJSON() ([]byte, error) {
 		if dep == nil {
 			continue
 		}
-		deps[repo] = Dependency{Sha: dep.Sha}
+		deps[repo] = Dependency{Sha: dep.Sha, GoVersionEnabled: dep.GoVersionEnabled}
 	}
 	return json.Marshal(deps)
 }


howardjohn:
@jasonwzm interested in this?

jasonwzm:
Does this work with the change? It did not work with my previous changes.

howardjohn:
@jasonwzm I am not 100% sure since we don't have a presubmit publish test, but I am pretty sure thisd will fix it

```
$ gsutil cat gs://istio-build/test/1.8.0-releasebuilder.745a484/manifest.yaml
dashboards:
  istio-mesh-dashboard: 7639
  istio-performance-dashboard: 11829
  istio-service-dashboard: 7636
  istio-workload-dashboard: 7630
  pilot-dashboard: 7645
dependencies:
  api:
    sha: 12ba196bf594cf81217aa128af4fe29c4ca29d36
  client-go:
    goversionenabled: true
    sha: b923a9337d297e46f8157b3d041c98434c7ea60a
```

before it would not keep the `goversionenabled`

jasonwzm:
ok, sg. Let's see how it goes.

PR #379, opened 2020-08-15
Remove ISTIO_VERSION variable

jwendell:
Not needed anymore since https://github.com/istio/common-files/pull/300


jwendell:
this needs to be merged first https://github.com/istio/common-files/pull/300

PR #395, opened 2020-08-28
Fix local build on linux

shamsher31:
`$(PWD)` breaks on Linux.  Instead, use `$(pwd)` which works fine on Mac and Linux . Also updated to latest image.

```bash
bash: PWD: command not found...
Similar command is: 'pwd'
docker: Error response from daemon: invalid mount config for type "bind": field Source must not be empty.
See 'docker run --help'.
```


ericvn:
I agree we can simplify it with @howardjohn's suggested changes. Another method that I probably use more often is to run `make shell` and run commands (same as John mentions) from within there.

PR #431, opened 2020-10-07
Re build 1.6.12

jacob-delgado:


brian-avery:
Holding off on approval for the moment. Reason it failed before was because the base image needed to be updated.

brian-avery (approved):

PR #458, opened 2020-10-28
Update envoy repo for 1.8

bianpengyuan:


bianpengyuan:
@ericvn Is this something that is not captured by branch cut automation?

bianpengyuan:
Actually looks like master has already switched to istio/envoy. So I guess this is just a one time update that does not need automation.

bianpengyuan:
Superseded by https://github.com/istio/release-builder/pull/457

ericvn:
@bianpengyuan I'll look into this but I suspect it shouldn't be an issue as we are no longer using envoy/envoy-wasm in master, and moved it to istio/envoy as well?

bianpengyuan:
Master is using envoyproxy/envoy now. I am not sure if that matters because the script only looks envoy repo for tagging during publishing, and we are never going to tag on master branch.

PR #460, opened 2020-10-30
Include tools/certs/common.mk

rvennam:
Fixes https://github.com/istio/istio/issues/28412


ericvn (approved):

ericvn:
I notice the file is in 1.8 as well, so this may need cherry-picking.

PR #505, opened 2020-12-14
Fix some directory issues in the container based build

ericvn:
Trying to run the build from within a container as directed in the readme fails in several areas. These changes get past the errors due to building files into different locations than copying them from.

I did this in two ways:
- Try copying from one of two locations
- Build into a location where the copy is expecting the output

The former is more difficult to do for the archives. I could try to do the later for the .rpm and .deb to be consistent.

### DIFF
--- a/pkg/build/archive.go
+++ b/pkg/build/archive.go
@@ -27,7 +27,10 @@ import (
 // istioctl, and various tools.
 func Archive(manifest model.Manifest) error {
 	// First, build all variants of istioctl (linux, osx, windows). gen-charts is required for manifests compiled in to istioctl.
-	if err := util.RunMake(manifest, "istio", nil, "gen-charts", "istioctl-all", "istioctl.completion"); err != nil {
+	// Generate the binaries in a specific directory which is used later to copy from.
+	// Oould use RepoOutDIr, but then needs to remomve the "/release" from the end.
+	env := []string{"TARGET_OUT=" + manifest.RepoDir("istio") + "/out/linux_amd64"}
+	if err := util.RunMake(manifest, "istio", env, "gen-charts", "istioctl-all", "istioctl.completion"); err != nil {
 		return fmt.Errorf("failed to make istioctl: %v", err)
 	}
 
--- a/pkg/build/archive.go
+++ b/pkg/build/archive.go
@@ -27,10 +27,7 @@ import (
 // istioctl, and various tools.
 func Archive(manifest model.Manifest) error {
 	// First, build all variants of istioctl (linux, osx, windows). gen-charts is required for manifests compiled in to istioctl.
-	// Generate the binaries in a specific directory which is used later to copy from.
-	// Oould use RepoOutDIr, but then needs to remomve the "/release" from the end.
-	env := []string{"TARGET_OUT=" + manifest.RepoDir("istio") + "/out/linux_amd64"}
-	if err := util.RunMake(manifest, "istio", env, "gen-charts", "istioctl-all", "istioctl.completion"); err != nil {
+	if err := util.RunMake(manifest, "istio", nil, "gen-charts", "istioctl-all", "istioctl.completion"); err != nil {
 		return fmt.Errorf("failed to make istioctl: %v", err)
 	}
 
--- a/pkg/build/debian.go
+++ b/pkg/build/debian.go
@@ -16,7 +16,6 @@ package build
 
 import (
 	"fmt"
-	"os"
 	"path"
 
 	"istio.io/release-builder/pkg/model"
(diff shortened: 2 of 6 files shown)


pkg/build/archive.go
```diff
@@ -27,7 +27,10 @@ import (
 // istioctl, and various tools.
 func Archive(manifest model.Manifest) error {
 	// First, build all variants of istioctl (linux, osx, windows). gen-charts is required for manifests compiled in to istioctl.
-	if err := util.RunMake(manifest, "istio", nil, "gen-charts", "istioctl-all", "istioctl.completion"); err != nil {
+	// Generate the binaries in a specific directory which is used later to copy from.
+	// Oould use RepoOutDIr, but then needs to remomve the "/release" from the end.
+	env := []string{"TARGET_OUT=" + manifest.RepoDir("istio") + "/out/linux_amd64"}
```
howardjohn, line 32: sorry, haven't been following this closely - what is wrong with the default output location? isn't it always going to be the same as this?

pkg/build/debian.go
```diff
@@ -27,9 +28,18 @@ func Debian(manifest model.Manifest) error {
 	if err := util.RunMake(manifest, "istio", nil, "deb/fpm"); err != nil {
 		return fmt.Errorf("failed to build sidecar.deb: %v", err)
 	}
-	if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
-		return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
+	if util.FileExists(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb")) {
+		if err := util.CopyFile(path.Join(manifest.RepoOutDir("istio"), "istio-sidecar.deb"), path.Join(manifest.OutDir(), "deb", "istio-sidecar.deb")); err != nil {
+			return fmt.Errorf("failed to package istio-sidecar.deb: %v", err)
+		}
+	} else if util.FileExists(path.Join(os.Getenv("TARGET_OUT_LINUX"), "release", "istio-sidecar.deb")) {
```
howardjohn, line 35: Where is TARGET_OUT_LINUX coming from? That variable is set per repo, so I am a bit confused why it would be set in release-builders environment?

pkg/build/archive.go
howardjohn: > /work/out/linux_amd64/release/istio-sidecar.deb which is where the file is created. But then the build tries to copy it from release/work/src/istio.io/istio/out/linux_amd64/release/istio-sidecar.deb

When we call `make deb` in istio/istio, we mount release/work/src/istio.io/istio as /work, So thse two paths should be identical.

howardjohn: I think in CI we have BUILD_WITH_CONTAINER=0 at the top level, so we have no docker containers. Then if a user users docker container for builder, the docker env sets some variables which tamper with the individual repo's variables?

ericvn: The problem with archive binary building on my Mac is that the binaries are built in the /tmp/istio-release/../darwin_amd64/ directory (as the container builds into the directory to match the host OS). If one was to search through all the potential build OS directories (and maybe architectures) in the few places we copy binaries, this adds a bunch of code. This change causes the make to always put the binaries in the linux output directory. It should be safe as the makefile is specifying the go build architecture when creating the files and the build container is running linux.

Running in a container on a Linux host wouldn't require this change and the build and copy from directories would both point to .../linx...

pkg/build/debian.go
ericvn: When built in a container, the files are built into /work/out/ (ex: /work/out/linux_amd64/release/istio-sidecar.deb). When built using prow, there are built in the tmp directories. We either need to check both places in the copy, or do as I did for archives, and always force the built artifacts to go to the /tmp directories.

pkg/build/archive.go
ericvn: IN my container:
```
env | grep OUT
CONTAINER_TARGET_OUT=/work/out/darwin_amd64
TARGET_OUT_LINUX=/work/out/linux_amd64
CONTAINER_TARGET_OUT_LINUX=/work/out/linux_amd64
TARGET_OUT=/work/out/darwin_amd64
```

pkg/build/debian.go
ericvn: Also, it appears in the logs that we build the .deb and .rpm in a few places. Before and after the images are built. It's the later copy that is failing.

ericvn: The docker portion of the build depends on the .deb/.rpm so they get built as part of that process. We build them again in the deb and rpm portions of the build and copy the files in the later portions.

ericvn:
@howardjohn When you have time for a review of my comments. Thanks.

uromahn:
@ericvn and @howardjohn Is there an ETA when this may be merged?

pkg/build/archive.go
ericvn: I tried a different approach, and maybe this is better.

The issues with copying artifacts when being run in a build container had to do with various `make` artifacts be created in the container's output directory as set by various env vars. Instead of trying to tweak the individual `make` jobs to generate the artifacts in the manifest specified directory, I am `unsetting` the container env vars. This better emulates what happens in the prow jobs that don't use a container.

pkg/build/debian.go
ericvn: See prior remark about unsetting env vars.

ericvn:
@uromahn I tweaked the design a little as noted in an earlier comment. I'll see if John likes this new approach better.

pkg/util/command.go
```diff
@@ -26,6 +26,16 @@ import (
 )
 
 func StandardEnv(manifest model.Manifest) []string {
+	// Unset those environment variables that are set in a container which causes `make xyz` artifacts
```
howardjohn, line 29: I agree with the design, but I don't think this function should be modifying state. I think we should either do it in an explicit function early in main(), modify `env` here to drop them (this impacts just a variable, not the real environment state), or have the doc that describes running in container drop them in the script before invoking

ericvn: #3 (update doc to have the `unset`s) is probably the easiest. Should update the doc to use `make shell` instead of calling the script directly (and the update to use `make shell` should be done regardless).
#2 is where I had hoped to go with the change. It's just easier to code up a few unset calls than to remove some map entries as a test.

ericvn: Update to only remove the env vars from the spawned `make` processes.

howardjohn (approved):

PR #507, opened 2020-12-24
Fix hub and tag for helm charts

jacob-delgado:
Current helm charts require the following to be passed in 

`    --set global.hub="docker.io/istio" --set global.tag=<version_to_upgrade>`

when installing in 1.8. In 1.9 this should be simplified by not specifying the hub and tag.

As a consequence, when testing (pre-release, qualification) we will have to set the hub appropriately to point to the pre-release image repository.


jacob-delgado:
cc @howardjohn @brian-avery @ericvn What are your thoughts on this? I tried to run `./release/build.sh`, but was having various issues that I think are done differently with prow.

I'm not sure about any other side-effects this would have. Also, the operator helm chart is not modified. If we decided to merge this, should we also do that as well?

pkg/build/archive.go
```diff
@@ -124,6 +127,30 @@ func Archive(manifest model.Manifest) error {
 	return nil
 }
 
+// sanitizeHelmCharts changes the hub and tag for all helm charts in archive, except
+// the operator
+func sanitizeHelmCharts(manifest model.Manifest, out string) error {
```
howardjohn, line 132: should we just find all **/values.yaml?

howardjohn (commented):
LGTM, can you add a test to pkg/validate/validate.go? There is already a similar one for operator, and if you go to old 1.4 history you can probably copy the old helm one as well

pkg/build/archive.go
jacob-delgado: I added back the capability in 1.4 which listed out the charts. It's not quite that elegant, but I figure it worked before. I can walk the tree if you think that's easier.

howardjohn (approved):

brian-avery (approved):

ericvn (approved):

PR #520, opened 2021-01-14
Use release-1.9 branch instead of master

jacob-delgado:
Change master to release-1.9
See https://github.com/istio/istio/issues/29490

### DIFF
--- a/release/build.sh
+++ b/release/build.sh
@@ -69,7 +69,7 @@ ${DEPENDENCIES:-$(cat <<EOD
     branch: release-1.9
   test-infra:
     git: https://github.com/istio/test-infra
-    branch: master
+    branch: release-1.9
   tools:
     git: https://github.com/istio/tools
     branch: release-1.9
--- a/release/build.sh
+++ b/release/build.sh
@@ -69,7 +69,7 @@ ${DEPENDENCIES:-$(cat <<EOD
     branch: release-1.9
   test-infra:
     git: https://github.com/istio/test-infra
-    branch: release-1.9
+    branch: master
   tools:
     git: https://github.com/istio/tools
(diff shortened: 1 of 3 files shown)


brian-avery (approved):

release/build.sh
```diff
@@ -62,17 +62,17 @@ ${DEPENDENCIES:-$(cat <<EOD
...
+    branch: release-1.9
     goversionenabled: true
   gogo-genproto:
     git: https://github.com/istio/gogo-genproto
-    branch: master
+    branch: release-1.9
   test-infra:
     git: https://github.com/istio/test-infra
-    branch: master
+    branch: release-1.9
```
howardjohn, line 72: this should be master  still, we don't branch test-infra

test/publish.sh
```diff
@@ -54,17 +54,17 @@ dependencies:
...
+    branch: release-1.9
     goversionenabled: true
   gogo-genproto:
     git: https://github.com/istio/gogo-genproto
-    branch: master
+    branch: release-1.9
   test-infra:
     git: https://github.com/istio/test-infra
-    branch: master
+    branch: release-1.9
```
howardjohn, line 64: same comment

howardjohn (changes requested):
Please remove my hold whenever, just want to make sure it doesn't get lost in the noise of 100s of PRs

test/publish.sh
jacob-delgado: Changed

release/build.sh
jacob-delgado: Changed

stevenctl (approved):

howardjohn (approved):

PR #557, opened 2021-02-04
do not ignore CVE vulns in base images

jacob-delgado:


jacob-delgado:
For https://github.com/istio/istio/issues/30601

release/build.sh
```diff
@@ -46,7 +46,6 @@ MANIFEST=$(cat <<EOF
 version: ${VERSION}
 docker: ${DOCKER_HUB}
 directory: ${WORK_DIR}
-ignoreVulnerability: true
```
shamsher31, line 49: can we just set it to false?

jacob-delgado: This is the way it was set originally https://github.com/istio/release-builder/pull/531/files

But I can change it

shamsher31: Thanks

shamsher31 (approved):

stevenctl (approved):

release/build.sh
ericvn: I'd have to go back and look, but I think the same script is run for all the commits in master (again, need to look), so in master they would ignore vulnerabilities, Maybe something to add to the release-branch scripts/list if that is the case.

PR #584, opened 2021-03-08
Add automated base image creation when base image scan fails

ericvn:


ericvn:
I believe this will need the `github` requirement added in `test-infra`

pkg/build/scanner.go
```diff
@@ -88,6 +89,50 @@ func Scanner(manifest model.Manifest) error {
...
+	}
+	newBaseVersion := baseVersion[:index] + strconv.Itoa(digit+1)
+	log.Infof("new baseVersion: %s", newBaseVersion)
+
+	// Attempted to run the ./tools/build-base-images.sh but locally, that gave me a
+	// number of different failures related to docker, PATHS, TTY, etc.
+	// Instead run the make command similar to what the script does.
+	buildEnv := []string{
+		"HUBS=docker.io/istio gcr.io/istio-release",
+		"TAG=" + newBaseVersion,
```
howardjohn, line 116: there is a small chance this could overwrite an existing image. Is it worth adding some check? `docker manifest inspect` can do it but not sure if there are better options. `docker manifest inspect` is also slow so we shouldn't put it in `dockerx.pushx` - but maybe `./tools/build-base-images.sh
`

pkg/build/scanner.go
```diff
@@ -88,6 +89,50 @@ func Scanner(manifest model.Manifest) error {
...
+		"BUILDX_BAKE_EXTRA_OPTIONS=--no-cache --pull",
+		"DOCKER_TARGETS=docker.base docker.distroless docker.app_sidecar_base_debian_9 docker.app_sidecar_base_debian_10 docker.app_sidecar_base_ubuntu_xenial docker.app_sidecar_base_ubuntu_bionic docker.app_sidecar_base_ubuntu_focal docker.app_sidecar_base_centos_7 docker.app_sidecar_base_centos_8",
```
howardjohn, line 118: this will make it a bit hard to maintain?

ericvn: Agreed and I preferred to use the script. The script just lead to a lot of different failures. I could go back to trying the script and all the various RunMake changes around the `cmd` again.

pkg/build/scanner.go
ericvn: I'll see if I can add it to the script.

pkg/build/scanner.go
ericvn: If I get it in the change to verify the images doesn't exist in the script, and can add a number of changes to actually get the script to work here, that should be good.

pkg/build/scanner.go
ericvn: Thoughts on checking if all images might already exist, or can one simply check istio/base?

howardjohn: probably just one is good enough? If its easier all of them is also fine, but it would be odd for that to happen

ericvn: https://github.com/istio/istio/pull/31323 was merged which iterates through and verifies we won't end up overwriting an image.

howardjohn (approved):

PR #623, opened 2021-04-14
Point build.sh back at istio/envoy

Monkeyanator:
Since we've [manually brought istio/envoy release-1.10 up to date with envoyproxy/envoy](https://github.com/istio/proxy/pull/3287) we can and should use `istio/envoy` for the builds now.

### DIFF
--- a/release/build.sh
+++ b/release/build.sh
@@ -76,7 +76,7 @@ ${DEPENDENCIES:-$(cat <<EOD
     git: https://github.com/istio/tools
     branch: release-1.10
   envoy:
-    git: https://github.com/envoyproxy/envoy
+    git: https://github.com/istio/envoy
     auto: proxy_workspace
 EOD
 )}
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

ericvn:
One will need to make sure that the istio/envoy release-1.10 branch has the SHA referenced in the istio/proxy WORKSPACE before doing another build. Today that is not the case because of a PR updating the SHA based on using envoyproxy/proxy to test a regression (without actually pulling the commits to istio/envoy).

brian-avery:
I think we still need to update https://github.com/istio/proxy/blob/release-1.10/WORKSPACE#L46 to point at the istio org.

brian-avery (approved):

brian-avery:
Opened proxy here: https://github.com/istio/proxy/pull/3296

Without this, I **think** automator will just try to update to the head of envoyproxy/envoy when this merges. With it, it should try to update to the head of release-1.10.

ZhiHanZ (approved):

PR #665, opened 2021-05-26
Trigger 1.10.1 build

Monkeyanator:


ericvn:
A quick scan this morning shows the `istio/base:1.10-dev.4` image has one vulnerability so the first run should create a new `istio/base:1.10-dev.5` which will need a manual PR updating the image before a successive build.

ZhiHanZ (approved):

PR #674, opened 2021-06-15
replace github commands with go-git

ericvn:
The current git code fails to push a commit to the upstream repo requesting a user id for the `https` linked repo. See https://storage.googleapis.com/istio-prow/logs/build-release_release-builder_release-1.10_postsubmit/1383997749321535488/build-log.txt for an example:
```
2021-04-19T04:19:53.704492Z	info	Running command: git push --set-upstream origin newBaseVersion1.10-dev.3
fatal: could not read Username for 'https://github.com': No such device or address
Error: failed to build: failed image scan: failed PR creation: exit status 128
```
To prevent having to specify the token on the commands, he code was rewritten to use a Golang git library. go-git is used as it will work on an already cloned filesystem with changes (which is what release-builder does). The other library used, google/go-github, works against an upstream repository. The later library is used to create the PR on the remote repository.

This change should fix the permission issue on the `git push` and the PR creation.

As a test, I
- copied a GitHub token to a local file, say ./etc/github-token/auth
- Ran `make shell`
- Within the shell, run `REPO_ORG=ericvn STEP=1 DRY_RUN=false GITHUB_TOKEN_FILE=/work/etc/github-token/oauth ./release/branch.sh`. This points the code to the GitHub token file (similar to how the pipeline add the token), runs code to create a change, and then creates a PR for that change.

Note that the failure above occurred when a base image failed during a build, and the new images were built, and the PR failed to be created. That scenario is something I didn't want to actually test, so used branching which is less complex/risky as I can override the GitHub org used by the code as the upstream repo.


pkg/util/git.go
```diff
@@ -15,107 +15,153 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to open path: %v", err)
+	}
+	w, err := r.Worktree()
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve work tree: %v", err)
 	}
-	if output.Len() == 0 {
+
+	// Git the worktree status to see if there are any changes. Return if none.
```
brian-avery, line 51: Get

ericvn: Fixed.

brian-avery (approved):

PR #675, opened 2021-06-16
Fix timeout retry and don't fail with error if ignoring vulnerabilties

ericvn:
Fixes: https://github.com/istio/istio/issues/32096

Fixed the current code to retry on timeouts. Also, if ignoreVulnerabilities is set, we should not fail on any scanner errors. Just log and continue.

### DIFF
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -63,8 +63,9 @@ func Scanner(manifest model.Manifest, githubToken string) error {
 		if err != nil {
 			if err, ok := err.(net.Error); ok && err.Timeout() {
 				fmt.Println("Scanner request timed out. Need to run request again.")
+			} else {
+				break
 			}
-			return err
 		}
 		numberRetries--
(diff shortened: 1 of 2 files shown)


pkg/build/scanner.go
```diff
@@ -63,8 +63,9 @@ func Scanner(manifest model.Manifest, githubToken string) error {
 		if err != nil {
 			if err, ok := err.(net.Error); ok && err.Timeout() {
 				fmt.Println("Scanner request timed out. Need to run request again.")
+			} else {
+				break
```
howardjohn, line 67: we never return the error now as far as I can tell

ericvn: err should be set and returned in the check immediately after the `for` loop

ericvn: That said, I can simply return the err here with the same behavior.

brian-avery (approved):

PR #687, opened 2021-07-02
Update to use Trivy

brian-avery:
This updates our release builder tooling to use trivy instead of IBM's image scanner. 

`ignoreVulnerability` is true, but a test run of this can be seen in the output of `dry-run_release-builder`

### DIFF
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -52,7 +52,7 @@ func Scanner(manifest model.Manifest, githubToken string) error {
 
 	// Call imagescanner passing in base image name. If request times out, retry the request
 	baseImageName := "istio/base:" + baseVersion
-	cmd := util.VerboseCommand("trivy", "--ignore-unfixed", "--exit-code", "2", baseImageName)
+	cmd := util.VerboseCommand("trivy", "--ignore-unfixed", "--no-progress", "--exit-code", "2", baseImageName)
 	if err = cmd.Run(); err != nil {
 		// There were either vulnerabilities or an issue running the scaner. Output message listing vulnerabilities.
 		log.Infof("Base image scan of %s failed with:\n %s", baseImageName, err.Error())
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -52,12 +52,10 @@ func Scanner(manifest model.Manifest, githubToken string) error {
 
 	// Call imagescanner passing in base image name. If request times out, retry the request
 	baseImageName := "istio/base:" + baseVersion
-	cmd := exec.Command("trivy", "--ignore-unfixed", baseImageName)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	cmd := util.VerboseCommand("trivy", "--ignore-unfixed", "--exit-code", "2", baseImageName)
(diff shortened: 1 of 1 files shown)


pkg/build/scanner.go
```diff
@@ -56,43 +52,11 @@ func Scanner(manifest model.Manifest, githubToken string) error {
...
-		resp, err = http.Get("http://imagescanner.cloud.ibm.com/scan?image=" + strings.TrimSpace(baseImageName))
-		if err != nil {
-			if err, ok := err.(net.Error); ok && err.Timeout() {
-				fmt.Println("Scanner request timed out. Need to run request again.")
-			} else {
-				return err
-			}
-		}
-		numberRetries--
+	cmd := util.VerboseCommand("trivy", "--ignore-unfixed", "--exit-code", "2", baseImageName)
```
ericvn, line 55: One issue is that privy is only part of the `master` build-tools at his point. If cherry-picking to the other releases, we need to add it there as well.

ericvn:
Also, is there some documentation, maybe from the PSWG, which states that there was some investigation into scanners and `Trivy` was the desired scanner? Just trying to prevent more PRs for additional scanners. I think we should have something documented somewhere.

brian-avery:
Agreed. 

It was a comparison between `docker scan`, `trivy` and the IBM scanner. Trivy and `docker scan` caught quite a bit more than the IBM scanner and Trivy was significantly faster with output that was quite a bit more readable. I also had some rate limiting issues with Docker which won't occur with Trivy given that it runs locally.

ericvn:
@brian-avery Would it be possible to get some confirmation from PSWG that `Trivy` is the tool going forward. I know you are the WG Lead, and if you say it is, that's fine. Just want some confirmation in the PR.

brian-avery:
Yep, Trivy is the tool we're moving to. 

@jacob-delgado Do you have any thoughts here?

pkg/build/scanner.go
ericvn, line 55: Can we also add `--no-progress` to suppress the progress bar:
```
2.98 MiB / 22.32 MiB [-------->_____________________________________________________] 13.37% ? p/s ?6.59 MiB / 22.32 MiB [------------------>___________________________________________] 29.54% ? p/s ?10.25 MiB / 22.32 MiB [---------------------------->________________________________] 45.93% ? p/s ?14.16 MiB / 22.32 MiB [------------------------------>_________________] 63.45% 18.65 MiB p/s ETA 0s18.67 MiB / 22.32 MiB [---------------------------------------->_______] 83.65% 18.65 MiB p/s ETA 0s22.32 MiB / 22.32 MiB [---------------------------------------------------] 100.00% 23.29 MiB p/s 1s2021-07-02T21:06:04.653Z	[34mINFO[0m
```

jacob-delgado:
> Yep, Trivy is the tool we're moving to. 
> 
> 
> 
> @jacob-delgado Do you have any thoughts here? 

This was talked about in the PSWG and was approved.

ericvn:
Thanks @jacob-delgado and @brian-avery. I'll approve after the one small change.

ericvn (approved):

PR #697, opened 2021-07-08
Add prow private job generation

brian-avery:
This will enable generation of the istio-private prow jobs during branch cutting. 

This needs to be merged after https://github.com/istio/test-infra/pull/3421 but I have tested it locally.

### DIFF
--- a/pkg/branch/setupProw.go
+++ b/pkg/branch/setupProw.go
@@ -33,6 +33,13 @@ func SetupProw(manifest model.Manifest, release string, dryrun bool) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("failed to generate new prow config: %v", err)
 	}
+
+	privateCmd := util.VerboseCommand("go", "run", "main.go", "branch", release)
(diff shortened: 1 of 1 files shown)


pkg/branch/setupProw.go
```diff
@@ -33,6 +33,13 @@ func SetupProw(manifest model.Manifest, release string, dryrun bool) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("failed to generate new prow config: %v", err)
 	}
+
```
ericvn, line 36: Update the comment above to say commands since we are adding another one.

ericvn (commented):
This particular step, #3, has been an issue to run locally for me as I don't have gcr.io permissions to retag images (comment is in the branch.go). But the addition of this additional command looks good.

pkg/branch/setupProw.go
brian-avery: Done.

ericvn (approved):

PR #706, opened 2021-07-14
Add RPM packages for CentOS 7 to the release

alexvanboxel:


alexvanboxel:
For whoever reviews this, we need the CentOS 7 RPM's, once approved how can this be backported to the 1.10 and 1.11 branch?

howardjohn:
I think this needs broader discussion on if the project will official support another platform, which has a testing, maintenance, etc cost.

alexvanboxel:
The binaries are already built, we've extracted them from a docker file now. Enterprises are not switching to CentOS 8 as it doesn't offer the long-term support that CentOS 7 delivers. Another reason why it should be supported is that the main reason to join the VM to a mesh is to gradually decompose a legacy system into microservices. If Istio is not going to support the last LTS type in the CentOS line why support VM's at all?

alexvanboxel:
It was one of the 1.10 milestones: https://github.com/istio/istio/milestone/24

alexvanboxel:
> I think this needs broader discussion on if the project will official support another platform, which has a testing, maintenance, etc cost.

How can we start this broader discussion? What's the right forum?

linsun:
@istio/wg-test-and-release-maintainers team, can you suggest how to move this forward for @alexvanboxel?  can we discuss this in an upcoming WG mtg?

linsun:
@alexvanboxel I recommend you consider joining one of the upcoming build/test/release WG mtg: https://github.com/istio/community/blob/master/WORKING-GROUPS.md#working-group-meetings

mikmatko:
> The binaries are already built, we've extracted them from a docker file now. **Enterprises are not switching to CentOS 8** as it doesn't offer the long-term support that CentOS 7 delivers.

Just came here to emphasize this. Any chance of this PR moving forward?

jacob-delgado:
> @istio/wg-test-and-release-maintainers team, can you suggest how to move this forward for @alexvanboxel? can we discuss this in an upcoming WG mtg?

cc @linsun @alexvanboxel 

I've added this to the T&R workgroup agenda for 9/28. Please attend to give your feedback. We are not committing to supporting CentOS 7 at this time, but in light of new information we can re-evaluate our position.

jwendell:
We have a misunderstanding here.

The current centos rpm package is actually made for centos 7.
See [this check](https://github.com/istio/proxy/blob/9437129a7ac9228c2a7bfc83a3929cb8af74f72f/scripts/release-binary.sh#L91) and the [base image](https://github.com/istio/tools/blob/823fd93b6b6611a931fad1e375a73231bdc0f4df/docker/build-tools/Dockerfile.centos#L1) where it is built. This image is used to build centos proxy [in prow](https://github.com/istio/test-infra/blob/8e4b362f3b2a33c4cbded9d2c1d6ccddb5d4903d/prow/config/jobs/proxy.yaml#L56).

So, from my understanding, nothing needs to be done.

jwendell:
hmm, never mind what I said. The envoy-centos binary is indeed for centos7 but the ordinary envoy works in centos8 and debian, and it's this one that's shipped. I got it.

So we are already building and testing envoy for centos7, only not shipping the resulting rpm in the release process. I don't think in any problem or additional cost in shipping it.

We could also rename current rpm to centos-8 so we end up shipping centos7 and centos8 rpm's, to make it more clear.

ericvn:
We did discuss this on today's T&R WG call. The author of this PR did not attend, but we did spend a fair amount of time discussing it. Some of the comments raised included:
- Concern about the older glibc and Envoy support long-term. 
- Not sure of current community commitment to continue to support another RPM long-term from the current attendees. Producing the image does imply support going forward for some amount of time. If there is an issue 1-yr down the road, who is going to look into it? A testing requirement was also mentioned. 
- istio-ecosystem might be a way to make it available and not put hardship on the community for long-term on-going support.

Some additional comments:
- Who will be users of the new RPM?
  -  Will we see a migration of CentOS 8 users back to CentOS 7 at the end of the year because they don't want to move to CentOS Stream? Or will they move to some other Linux?
  - Will there be new users if CentOS 7 is going away at the end of 2024?
  - This might be more of a question for the Environments WG. What is seen as a requirement going forward?

Some unanswered questions from the Istio standpoint. If CentOS 8 is no longer supported, do we need to continue support the CentOS 8 RPM, or should we announce support ending as well.
- One thought would be to switch from providing a CentOS 8 RPM instead to a CentOS 7 RPM. But that still leads to the questions above.

PR #707, opened 2021-07-14
Add return for successful scan

brian-avery:
I'd like to refactor this and move the PR creation into its own function, but for now this returns on success and finds non-vulnerability errors and returns. For everything else, it continues on.


pkg/build/scanner.go
```diff
@@ -53,12 +53,24 @@ func Scanner(manifest model.Manifest, githubToken string) error {
...
+	err = cmd.Run()
+	if err == nil {
+		log.Infof("Base image scan of %s was successful", baseImageName)
+		return nil
+	}
+
+	//--exit-code 2 above states to return 2 if vulnerabilities are found. If we get a different error code or we can't check the error code, bail out
+	if exitError, ok := err.(*exec.ExitError); ok {
+		// Scanner failed with an exit code indicating a failure other than vulnerabilities found
+		if exitError.ExitCode() != -2 {
```
ericvn, line 65: Should this be a compare to 2?

ericvn (approved):

PR #718, opened 2021-07-15
Fix issue where manifest dependencies are SHAs during build

ericvn:
Looking at the release-builder failure: 

https://storage.googleapis.com/istio-prow/logs/build-release_release-builder_release-1.10_postsubmit/1415393564543684608/build-log.txt

There are 2 issues: The first is that the `trivy` command found no issues so we shouldn't have tried building new images and create the PR for them. This problem was fixed in  https://github.com/istio/release-builder/pull/707.

The second issue was the failing `panic: runtime error: index out of range [-1]`

Looking at the code, I see that we attempt to use the `istio` manifest dependency and get the `git url` and `branch`to create the PR. This fails in the case of the build because they were overriden by the `StandardizeManifest` call replacing them with the SHA of the head of the git url and branch.

The new code saves the git url and branch from the input manifest, and passes them along as parameters. In the `createPR` method we check if they are blank (this happens in the case of branching), and then set them from the manifest (branching does not Standardize the manifest).

I did move the Scanner call to the command from the build only to prevent having to pass the new parts on the build call.

### DIFF
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -38,7 +38,7 @@ type Results struct {
 }
 
 // Scanner checks the base image for any CVEs.
-func Scanner(manifest model.Manifest, githubToken string) error {
+func Scanner(manifest model.Manifest, githubToken, git, branch string) error {
 	// Retrieve BASE_VERSION from the istio/istio Makefile
 	istioDir := manifest.RepoDir("istio")
 	var out bytes.Buffer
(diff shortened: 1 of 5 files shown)


pkg/build/scanner.go
```diff
@@ -96,21 +96,21 @@ func Scanner(manifest model.Manifest, githubToken string) error {
...
 	sedCmd := util.VerboseCommand("sed", "-i", sedString, "Makefile.core.mk")
-	sedCmd.Dir = manifest.RepoDir("istio")
+	sedCmd.Dir = istioDir
 	if err := sedCmd.Run(); err != nil {
 		return fmt.Errorf("failed to run sed command: %v", err)
 	}
 
 	if err := util.CreatePR(manifest, "istio", "newBaseVersion"+newBaseVersion,
-		"Update BASE_VERSION to "+newBaseVersion, false, githubToken); err != nil {
+		"Update BASE_VERSION to "+newBaseVersion, false, githubToken, git, branch); err != nil {
```
Monkeyanator, line 113: nit: maybe we can make the args a struct to make it more readable

Monkeyanator (approved):

PR #723, opened 2021-07-15
[release-1.9] Manual cherry-pick of #721

ericvn:
Fix issue where manifest dependencies are SHAs during build

This should allow the PR to be created when new base images are built so you don't have to. Just approve the one created.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -31,16 +31,6 @@ import (
 // Build will create all artifacts required by the manifest
 // This assumes the working directory has been setup and sources resolved.
 func Build(manifest model.Manifest, githubToken string) error {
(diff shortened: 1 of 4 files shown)


shamsher31 (commented):
Why is this file missing https://github.com/istio/release-builder/pull/718/files#diff-6a3ac39c79d2345b667d1de3073e1038ae0b33767751ea4c85dda724d61aa686R95 ?

ericvn:
@shamsher31 The `branch` command was never part of release 1.9, starting in the 1.10 branch. No reason to cherrypick the `branch` command back as we won't be branching from that release in the future. Thus the file never existed in 1.9 and the reason I had to do the cherrypick manually.

jacob-delgado (approved):

PR #758, opened 2021-08-16
Rewrite appVersion

howardjohn:
Currently our helm charts do not set this; I think they should to follow
helm standards. This is the first step.


jacob-delgado (approved):
appVersion is considered optional, but this is fine to include.

PR #761, opened 2021-08-20
Sanitize istiod-remote charts

frankbu:
Fixes: https://github.com/istio/istio/issues/34804


ericvn:
Might it be better to generate a list of charts using something like: `find ./manifests -name Chart.yaml`

howardjohn (approved):

PR #778, opened 2021-09-10
Create 1.12.0-alpha.0

howardjohn:
Get the first alpha build started. Mostly to test helm repo


release/trigger-build
```diff
@@ -1 +1 @@
-1.10.0-alpha.0
```
mandarjog, line 1: Why a jump from 1.10 to 1.12 ?
Did we not follow the normal process before?

mandarjog (approved):
Lgtm

ericvn:
A base image with vulnerabilities was detected, the new base images look to have been created and a PR to use them was as well. I've approved that PR so another spin should get the build going.

release/trigger-build
howardjohn: Yeah we forgot to do a 1.11 alpha

PR #790, opened 2021-09-21
Add cosign image signing support to release-builder

stewartbutler:
Adds support for signing images to the release-builder workflow. Depends on deployment of https://github.com/istio/test-infra/pull/3573

### DIFF
--- a/pkg/publish/cmd.go
+++ b/pkg/publish/cmd.go
@@ -40,6 +40,7 @@ var (
 		github       string
 		githubtoken  string
 		grafanatoken string
+		cosignkey    string
 	}{}
 	publishCmd = &cobra.Command{
 		Use:          "publish",
@@ -84,6 +85,8 @@ func init() {
 		"The file containing a github token.")
 	publishCmd.PersistentFlags().StringVar(&flags.grafanatoken, "grafanatoken", flags.grafanatoken,
 		"The file containing a grafana.com API token.")
+	publishCmd.PersistentFlags().StringVar(&flags.cosignkey, "cosignkey", flags.cosignkey,
+		"A key for signing images, as passed to cosign using 'cosign sign -key <x>'")
 }
 
 func GetPublishCommand() *cobra.Command {
@@ -99,7 +102,7 @@ func validateFlags() error {
 
 func Publish(manifest model.Manifest) error {
 	if flags.dockerhub != "" {
-		if err := Docker(manifest, flags.dockerhub, flags.dockertags); err != nil {
+		if err := Docker(manifest, flags.dockerhub, flags.dockertags, flags.cosignkey); err != nil {
 			return fmt.Errorf("failed to publish to docker: %v", err)
 		}
 	}
--- a/pkg/publish/docker.go
+++ b/pkg/publish/docker.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"istio.io/pkg/log"
 	"istio.io/release-builder/pkg/model"
 	"istio.io/release-builder/pkg/util"
 )
@@ -36,7 +37,9 @@ func Docker(manifest model.Manifest, hub string, tags []string, cosignkey string
 	// able to run 'cosign public-key <key>'.
 	cosignEnabled := false
 	if cosignkey != "" {
-		if err := util.VerboseCommand("cosign", "public-key", "-key", cosignkey).Run(); err == nil {
+		if err := util.VerboseCommand("cosign", "public-key", "-key", cosignkey).Run(); err != nil {
+			log.Errorf("Argument '--cosignkey' nonempty but unable to access key %v, disabling signing.", err)
+		} else {
 			cosignEnabled = true
 		}
 	}
--- a/pkg/publish/docker.go
+++ b/pkg/publish/docker.go
@@ -26,10 +26,19 @@ import (
 )
 
 // Docker publishes all images to the given hub
-func Docker(manifest model.Manifest, hub string, tags []string) error {
+func Docker(manifest model.Manifest, hub string, tags []string, cosignkey string) error {
 	dockerArchives, err := ioutil.ReadDir(path.Join(manifest.Directory, "docker"))
 	if err != nil {
 		return fmt.Errorf("failed to read docker output of release: %v", err)
+	}
+
+	// Only attempt to sign images if a valid cosign key is provided and we are
+	// able to run 'cosign public-key <key>'.
+	cosignEnabled := false
+	if cosignkey != "" {
+		if err := util.VerboseCommand("cosign", "public-key", "-key", cosignkey).Run(); err == nil {
+			cosignEnabled = true
+		}
 	}
 
 	for _, f := range dockerArchives {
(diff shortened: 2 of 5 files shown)


pkg/publish/cmd.go
```diff
@@ -84,6 +85,8 @@ func init() {
 		"The file containing a github token.")
 	publishCmd.PersistentFlags().StringVar(&flags.grafanatoken, "grafanatoken", flags.grafanatoken,
 		"The file containing a grafana.com API token.")
+	publishCmd.PersistentFlags().StringVar(&flags.cosignkey, "cosignkey", flags.cosignkey,
```
howardjohn, line 88: for github token we use a file. Should we do similar here? or read from env?

howardjohn: I guess if `cosign` takes it as a flag so can we

test/publish.sh
```diff
@@ -86,6 +87,16 @@ go run main.go build --manifest <(echo "${MANIFEST}")
...
 go run main.go validate --release "${WORK_DIR}/out"
 
+if [[ -z "${COSIGN_KEY}" ]]; then
+  COSIGN_ARGS="--cosignkey ${COSIGN_KEY}"
+fi
+
 if [[ -z "${DRY_RUN:-}" ]]; then
-  go run main.go publish --release "${WORK_DIR}/out" --helmbucket "${HELM_BUCKET}" --gcsbucket "${GCS_BUCKET}" --dockerhub "${DOCKER_HUB}" --dockertags "${VERSION}"
+go run main.go publish --release "${WORK_DIR}/out" \
+  "${COSIGN_ARGS:-}" \
```
howardjohn, line 96: does this work? I thought this would make a single arg and we want two?

```suggestion
  --cosignkey "${COSIGN_KEY:-}" \
```

May be simpler and give the same behavior

howardjohn (approved):

pkg/publish/cmd.go
jwendell: +1 for using a file. Using env is dangerous as we can leak its content as it's passed as CLI arguments to commands

pkg/publish/docker.go
```diff
@@ -26,12 +26,21 @@ import (
...
 		return fmt.Errorf("failed to read docker output of release: %v", err)
 	}
 
+	// Only attempt to sign images if a valid cosign key is provided and we are
+	// able to run 'cosign public-key <key>'.
+	cosignEnabled := false
+	if cosignkey != "" {
+		if err := util.VerboseCommand("cosign", "public-key", "-key", cosignkey).Run(); err == nil {
+			cosignEnabled = true
+		}
```
jwendell, line 41: Let's print the error here. `cosign` key was passed, user expects it to be used.

pkg/publish/cmd.go
ericvn: Looking at the cosign command, `-key` is `path to the private key file, KMS URI or Kubernetes Secret` so not the actual key. So as discussed in the T&R WG call, the path will end up being output in the `util.VerboseCommand`, and not the actual key. I would like to have Path in the variable name so we understand that, but cosign uses the `-key` flag.

pkg/publish/docker.go
ericvn: Is the point here to see the returned `public-key` or just to verify that the command exists? If the later, does it make sense to just do a `-version`?

test/publish.sh
ericvn: While I like John's suggestion, I suspect that it will result in a `flag needs an argument:` error.

howardjohn: `--cosignkey "" --githubtoken ...` is different from `--cosignkey --githubtoken ...` I believe. Not 100% sure though

stewartbutler: Right, the intention for the `${COSIGN_ARGS:-}` placeholder is to hold the entire `--cosignkey gcpkms://path/to/whatever` argument, or skip it with an empty string if no `${COSIGN_KEY}` is present. That means that if a job uses the publish workflow but does not include the `${COSIGN_KEY}` env variable, it will still run, making it easier to migrate without breaking people.

pkg/publish/docker.go
stewartbutler: It does both. If a key is passed but the `cosign` binary is not present, it will abort attempting to sign the images without failure. If the cosign binary is present but the key is invalid or the permissions or wrong, it will also abort without a failure.

It wasn't clear to me if we wanted to return an error in that case and exit or if we should log a warning and continue.

pkg/publish/cmd.go
stewartbutler: Right, the expected value here (in the case of Prow) is a GCP KMS key of format `gcpkms://projects/host-project/locations/global/keyRings/keyring-name/cryptoKeys/key-name/versions/1`, as documented in https://github.com/sigstore/cosign/blob/main/KMS.md. That *path* is stored in the prow cluster in a secret, and is exposed as an environment variable via the kubernetes secrets api in istio/test-infra/pull/3573.

If you guys think it would be better to instead expose it as a file and then read it in the go code I can do that as well.

test/publish.sh
howardjohn: I get the intention I just don't think it works?
```
$ x="-l -a"
$ ls "${x}
ls: invalid option -- ' '
Try 'ls --help' for more information.
```

stewartbutler: I quoted badly -- to do this trick you have to unquote it.
```
$ cat test.sh
───────┬────────────────────────────────────────────────────────────────────
       │ File: test.sh
───────┼────────────────────────────────────────────────────────────────────
   1   │ #!/bin/bash
   2   │ 
   3   │ 
   4   │ 
   5   │ echo "Testing with TESTARG=${TESTARG}"
   6   │ TESTARG="-l -a"
   7   │ ls \
   8   │   ${TESTARG:-} \
   9   │   ./
  10   │ 
  11   │ echo "Testing with TESTARG unset"
  12   │ unset TESTARG
  13   │ ls \
  14   │   ${TESTARG:-} \
  15   │   ./
───────┴────────────────────────────────────────────────────────────────────
$ ./test.sh  
Testing with TESTARG=
total 12
drwxr-x---   2 stewartbutler primarygroup 4096 Sep 21 13:14 .
drwxr-x--- 155 stewartbutler primarygroup 4096 Sep 21 13:13 ..
-rwxr-x---   1 stewartbutler primarygroup  173 Sep 21 13:18 test.sh
Testing with TESTARG unset
test.sh
```

ericvn: > `--cosignkey "" --githubtoken ...` is different from `--cosignkey --githubtoken ...` I believe. Not 100% sure though

I  missed the quotes. I do believe what you suggested John should work fine.

pkg/publish/docker.go
ericvn: I'm fine logging a warning and continuing if `cosign`ing an image is an optional step. If we think the step is really required, then we should log and fail.

test/publish.sh
stewartbutler: Switched to John's idea -- after some testing it looks like the arg parsing is smart enough to deal with empty quotes.

pkg/publish/docker.go
stewartbutler: Added a warning log.

johnma14 (approved):

PR #793, opened 2021-09-24
Skip helm charts for CI builds

sergii-ssh:
In CI jobs helm chart version semantics differ from what is used for releases and that causes failures in a build flow. Hence skipping helm charts for non release builds

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -38,10 +38,10 @@ func Build(manifest model.Manifest, githubToken string) error {
 		}
 	}
 
+	if err := SanitizeAllCharts(manifest); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/build/build.go
```diff
@@ -37,13 +38,17 @@ func Build(manifest model.Manifest, githubToken string) error {
 		}
 	}
 
-	if err := SanitizeAllCharts(manifest); err != nil {
```
howardjohn, line 40: I think we want this all of the time. We still want to local helm chart files sanitized, just skip the step to package them with `helm package`.

So just HelmCharts in the 'if'

sergii-ssh: done

howardjohn (approved):

PR #796, opened 2021-09-29
Updates for new tagging scheme

howardjohn:
Blocking for https://github.com/istio/istio/pull/35351/

### DIFF
--- a/pkg/publish/docker.go
+++ b/pkg/publish/docker.go
@@ -26,6 +26,8 @@ import (
 	"istio.io/release-builder/pkg/util"
 )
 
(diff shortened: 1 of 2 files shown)


howardjohn:
There is a flaw in our testing that its impossible to change the output files of the release since both istio and release-builder need to change atomically.
We have two options:
* Make the test allow both forms, merge in istio, then make it allow one form
* Just force merge over the test failure.

Slightly prefer the latter to save some time but I can do (1) if there are concerns

jwendell (approved):
I'm fine with force merge

PR #800, opened 2021-09-30
Fix build release tag

howardjohn:
turns out the old idea didn't work. Unfortunately was a bit overconfident and force merged the old one...


pkg/build/docker.go
```diff
@@ -26,7 +26,7 @@ import (
 // docker.save in the repos does most of the work, we just need to call this and copy the files over
 func Docker(manifest model.Manifest) error {
 	// Build both default and distroless variants
```
Monkeyanator, line 28: nit: update comment?

Monkeyanator (approved):

PR #801, opened 2021-10-01
Skip validate when version is not semver

howardjohn:
Also make the check identical to helm's logic


nmittler (approved):

howardjohn:
Force merging since we have a codependency on istio/istio getting this PR for the test to pass

PR #821, opened 2021-10-20
Set default revision tag to default for fresh installs

jacob-delgado:
Fixes Item 1 in https://github.com/istio/istio/issues/35690

This will NOT change anything for istioctl based installs, but will only change helm charts before they are published. This will allow the users to not have to set `--set defaultRevision=default` on fresh installs.

### DIFF
--- a/pkg/build/helm.go
+++ b/pkg/build/helm.go
@@ -140,7 +140,7 @@ func sanitizeChart(manifest model.Manifest, s string) error {
 				contents = strings.ReplaceAll(contents, before, after)
 			}
 
-			if filepath.Base(s) == path.Join(manifest.RepoDir("istio"), baseChart) && fname == "values.yaml" {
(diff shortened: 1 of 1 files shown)


pkg/build/helm.go
```diff
@@ -137,6 +140,12 @@ func sanitizeChart(manifest model.Manifest, s string) error {
 				contents = strings.ReplaceAll(contents, before, after)
 			}
 
+			if s == path.Join(manifest.RepoDir("istio"), baseChart) && fname == "values.yaml" {
+				before := `defaultRevision: ""`
+				after := "defaultRevision: default"
```
howardjohn, line 145: why does this happen in this repo instead of values.yaml in istio/istio?

jacob-delgado: After a little bit more tracking this change was done when the validating webhook was moved from discovery to base. I'd prefer we go with that change. Let me experiment with that as it'd be closer to where we'd want it to live.

PR #832, opened 2021-11-02
Add version prefix to base image tag

Monkeyanator:
Looked into adding the `:latest` support too... I think what we want is to check if we're building off master branch and if we are also tag as latest but not immediately clear to me how to check which branch the automation is running against.

### DIFF
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -73,8 +73,8 @@ func Scanner(manifest model.Manifest, githubToken, git, branch string) error {
 	// Time format chosen for consistency with build tools tag:
 	// https://github.com/istio/tools/blob/ee7da00900dc878a2e865e43250c34735f130b7a/docker/build-tools/build-and-push.sh#L27
 	const timeFormat = "2006-01-02T15-04-05"
-	buildTimestamp := time.Now().Format(timeFormat)
-	log.Infof("new base tag: %s", buildTimestamp)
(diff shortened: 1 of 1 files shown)


jacob-delgado (commented):
cherry pick back to other releases?

Monkeyanator:
> cherry pick back to other releases?

Good call, think these changes are only on 1.12 right now so cherry-picking there.

pkg/build/scanner.go
```diff
@@ -73,8 +73,8 @@ func Scanner(manifest model.Manifest, githubToken, git, branch string) error {
 	// Time format chosen for consistency with build tools tag:
 	// https://github.com/istio/tools/blob/ee7da00900dc878a2e865e43250c34735f130b7a/docker/build-tools/build-and-push.sh#L27
 	const timeFormat = "2006-01-02T15-04-05"
-	buildTimestamp := time.Now().Format(timeFormat)
-	log.Infof("new base tag: %s", buildTimestamp)
+	tag := fmt.Sprintf("%s-%s", manifest.Version, time.Now().Format(timeFormat))
```
howardjohn, line 76: nit: the build-tools uses the actual branch, not version. Version is slightly different.

I am not against this, but if we already have to figure out the branch for `:latest` I slightly prefer using thebranch for consistency

howardjohn (approved):

PR #847, opened 2021-11-15
Trigger 1.10.6 build

Monkeyanator:


ericvn:
First pass should create new base images (trivy fails vulnerability check (although the old image scanner didn't)).

ZhiHanZ (approved):

PR #890, opened 2022-01-14
Docker/allow load

howardjohn:
Some build systems (eg the one we use, unfortunately :-) ), do not fair
too well with targz output. We can save them + compress + uncompress +
load, but its a lot of wasted efforts. The old docker builder used to
save *and* load, but with buildx you can only have a single output for
now, so making this an explicit option.

This does not impact anyone unless they change their manifest
explicitly.

I also updated go modules.

cc @sergii-ssh


ericvn:
LGTM. Will wait on a response from the cc.

ericvn (approved):
John has already had a discussion with Sergei, so approving.

sergii-ssh:
thank you John and Eric

PR #897, opened 2022-01-20
fix prow gen path

stevenctl:
Matches changes in https://github.com/istio/test-infra/pull/3771/


stevenctl:
seems like the command still isn't working.. now it has a nil pointer issue

stevenctl:
should be good to go

ericvn (approved):

PR #903, opened 2022-01-24
helm: publish to OCI registry

howardjohn:
Fixes https://github.com/istio/istio/issues/36976

Users will be able to install like `helm template
oci://gcr.io/istio-release/charts/base`


jacob-delgado (approved):
lgtm. I didn't know this existed or was in-flight.

PR #963, opened 2022-02-28
Converting SBOM error into warning -- lack of BOM should not be fatal.

stewartbutler:
One more release builder tweak to SBOM.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -89,8 +89,14 @@ func Build(manifest model.Manifest) error {
 		return fmt.Errorf("failed to package license file: %v", err)
 	}
 
-	if err := GenerateBillOfMaterials(manifest); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/build/build.go
```diff
@@ -90,7 +90,7 @@ func Build(manifest model.Manifest) error {
 	}
 
 	if err := GenerateBillOfMaterials(manifest); err != nil {
-		return fmt.Errorf("failed to generate sbom: %v", err)
+		log.Warnf("failed to generate sbom: %v", err)
```
howardjohn, line 93: Seems like it should be fatal? if we don't want it in some contexts we should explicitly disable?

stewartbutler: Adding a flag for it instead.

johnma14 (approved):
LGTM

PR #990, opened 2022-03-10
Debug why helm repo index never works

howardjohn:
1.13.2 failed to publish again.

Also `make test` never does anything, so I fixed it to be phony


jacob-delgado:
Should we add a check that does an helm update or http get? The job shouldn't fail, but we need to make it part of the release cycle. Just not sure how to do it effectively.

jacob-delgado (approved):

PR #995, opened 2022-03-15
Generate the license tar for release-builder

ericvn:
Fixes: https://github.com/istio/istio/issues/37909

See issue for some discussion on ways to accomplish this.

Settled on adding a new auto method to get the SHA from the istio/istio prow/release-commit.sh file and specifying that in the manifests.


example/manifest.yaml
```diff
@@ -51,6 +51,9 @@ dependencies:
   envoy:
     git: https://github.com/envoyproxy/envoy
     auto: proxy_workspace
+  release-builder:
+    git: https://github.com/istio/release-builder
+    auto: release-commit
```
howardjohn, line 56: a bit confused by this. In istio/release-builder, we should not read from istio/istio

In istio/istio, we already have BUILDER_SHA in the same file that we create the manifest, so why not just reference the `sha: BUILDER_SHA`?

ericvn: @howardjohn We already read the proxy SHA from istio/istio (from istio.deps when using auto:deps). This extends that behavior to allow the reading of the release-builder SHA being used in the release-commit.sh file.

ericvn: I suppose we could change to instead of reading from istio/istio, specify the specific sha in that file, removing the new code. Then we can just specify the branches in the manifests here as using the last commit would be fine.

howardjohn: We actually _build_ the proxy from that SHA though. release-commit.sh is strictly used in istio/istio presubmit/postsubmit tests. It is not used at all in the real release build

ericvn: So, then I fall back to option 1 or 2? Does it make sense to add another dependency for release-builder to get placed in the source directory to get the license files, or should we use the code that was already cloned via the pipeline? IN other words, create yet another source directory that would be found in the range of dependencies to grab from licenses, or just copy the licenses from the current release-builder?

howardjohn (approved):

PR #1004, opened 2022-03-23
Add multi platforms support for RepoOutDir

TrevorTaoARM:
Add multi-platform output dir support for
arm64 and other platforms, which is used in
archive, package(deb, rpm), docker
build actions.

Signed-off-by: TrevorTaoARM <trevor.tao@arm.com>

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -16,7 +16,9 @@ package model
 
 import (
 	"encoding/json"
+	"fmt"
 	"path"
+	"runtime"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
I am fine doing this short term but I think for Istio official builds we would probably only be doing both ARM and AMD at once, so we may need more work there in that regard

howardjohn:
Actually this would mean that users on non-linux/amd64 can no longer build a proper release. For example, someone running on OSX now gets a osx build which is not what we want. The build platform != target platform IMO

ericvn:
Doing some testing with MacOS as John mentioned, but running into another issue that I need to fix first :(.

PR #1127, opened 2022-07-18
Trigger publish of release 1.14.2

lei-tang:
Trigger publish of release 1.14.2.


GregHanson (approved):

sathieu:
@GregHanson @lei-tang Something went wrong with this release? There is no assets in: https://github.com/istio/istio/releases/tag/1.14.2

ericvn:
@sathieu As the release has not been officially released yet, I don't think the assets should be available. I believe (by their channel and the release notes) that the release should be available on the 25th. The actual publish here should not have happened until then.

sathieu:
Thanks for the info @ericvn.

PR #1172, opened 2022-08-17
trigger 1.15.0-beta.1 build

dhawton:


ericvn:
Just added a DNM to keep the PRs from merging when approved (as publish likely will happen later). 

Also note that an update_deps.sh can be run before the build to make sure a few of dependencies get updated.

Monkeyanator (approved):

kzdv-pipeline:
> Just added a DNM to keep the PRs from merging when approved (as publish likely will happen later).
> 
> Also note that an update_deps.sh can be run before the build to make sure a few of dependencies get updated.

Thanks Eric. Done and PR opened https://github.com/istio/istio/pull/40529. Will wait for that to merge before removing the label here.

PR #1219, opened 2022-10-06
Trigger the build for Istio 1.13.9

lei-tang:
Trigger the build for Istio 1.13.9


stevenctl (approved):

ericvn:
Merged without CLA check since that appears to be an outstanding issue right now.

PR #1220, opened 2022-10-06
Trigger the publish of Istio 1.13.9

lei-tang:
Trigger the publish of Istio 1.13.9.


lei-tang:
A do-not-merge label is added to publish the build only upon release.

PR #1237, opened 2022-10-17
Test - Add semver err to log

ericvn:
Also updated common. Investigating failure


ericvn:
Let's merge this one as a supersedes as this contains th common-files update plus some additional messaging which might be useful later.

howardjohn (approved):

PR #1254, opened 2022-10-24
fix-publish-vars

jtischer-acasi:
replaced hard coded helm & dockerhub


release/publish.sh
```diff
@@ -66,6 +67,6 @@ go run main.go publish --release "${WORK_DIR}" \
 # Docker hub doesn't support Helm registries, so we also push these only to GCR.
 go run main.go publish --release "${WORK_DIR}" \
     --cosignkey "${COSIGN_KEY:-}" \
-    --helmhub "gcr.io/istio-release/charts" \
-    --dockerhub "gcr.io/istio-release" \
+    --helmhub "gcr.io/${HELM_BUCKET}" \
```
howardjohn, line 70: can you make this its own var? The fact the HELM_BUCKET _happens_ to also be the gcr.io relative URL is purely coincidence

howardjohn (commented):
I am not against this.. but publish.sh isn't really intended to be generic. I'd recommend vendors building have their own script (you'll note that 95% of the script is just user-specific variables)

howardjohn (approved):

PR #1294, opened 2022-12-07
Add warning notes if helm charts Istio version is already EOL

zufardhiyaulhaq:
Signed-off-by: zufardhiyaulhaq <zufardhiyaulhaq@gmail.com>

fix https://github.com/istio/istio/issues/41747, introduce a warning if the current date is near or after EOL of the date defined in the release builder.

image

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -131,6 +131,9 @@ type InputManifest struct {
 	Dependencies IstioDependencies `json:"dependencies"`
 	// Version specifies what version of Istio this release is
 	Version string `json:"version"`
+	// EOLDate specifies what is the date when this Istio version become EOL
+	// Should be in format "yyyy-mm"
+	EOLDate string `json:"eolDate"`
 	// Docker specifies the docker hub to use in the helm charts.
 	Docker string `json:"docker"`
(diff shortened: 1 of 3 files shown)


pkg/model/model.go
```diff
@@ -160,6 +163,9 @@ type Manifest struct {
 	Dependencies IstioDependencies `json:"dependencies"`
 	// Version specifies what version of Istio this release is
 	Version string `json:"version"`
+	// EOLDate specifies what is the date when this Istio version become EOL
+	// Should be in format "yyyy-mm"
+	EOLDate string `json:"eolDate"`
```
howardjohn, line 168: If we already have to update istioctl binary directly it seems like we could just update the helm notes directly as well? We can have a test to check they are in sync if needed

Will defer to @ericvn

zufardhiyaulhaq: do you mean to update the helm notes directly on istio/istio repository?

zufardhiyaulhaq:
@ericvn can you help here?

zufardhiyaulhaq:
cc @istio/wg-test-and-release-maintainers please help review

kfaseela:
cc @jacob-delgado

PR #1316, opened 2022-12-19
Updating build.sh and and test/publish.sh

jtischer-acasi:
- test/publish.sh - missing improvements from build.sh so it can utilize the same ENV variables for dependancies and PROXY_OVERRIDE
- build.sh - changed hardcoded github repo urls to use GITHUB_ORG


test/publish.sh
```diff
@@ -27,55 +27,66 @@ else
...
 
-DOCKER_HUB=${DOCKER_HUB:-gcr.io/istio-testing}
-GCS_BUCKET=${GCS_BUCKET:-istio-build/test}
-HELM_BUCKET=${HELM_BUCKET:-istio-build/test/charts}
-VERSION="1.14.0-releasebuilder.$(git rev-parse --short HEAD)"
+GCS_BUCKET=${GCS_BUCKET:-istio-prerelease/prerelease}
+HELM_BUCKET=${HELM_BUCKET:-istio-prerelease/charts}
 COSIGN_KEY=${COSIGN_KEY:-}
+GITHUB_ORG=${GITHUB_ORG:-istio}
+DOCKER_HUB=${DOCKER_HUB:-docker.io/istio}
```
Kmoneal, line 34: Is there any particular reason this is switched from `gcr.io` to `docker.io`?

jjtischer: the build.sh script is a presubmit job and is set to "DOCKER_HUB=${DOCKER_HUB:-docker.io/istio}" so my reasoning is that the test/publish.sh job is a postsubmit job and it seemed to me that this test should validate what was created from the build job.

Kmoneal (approved):

test/publish.sh
howardjohn: this is the rest of publish, not the real publish. We don't want to go to the real dockerhub

jjtischer: reverted DOCKER_HUB

ericvn (changes requested):
And not sure why these changes are in 1.16 and not the main branch and cherry-picked.

test/publish.sh
```diff
@@ -27,55 +27,66 @@ else
...
-VERSION="1.14.0-releasebuilder.$(git rev-parse --short HEAD)"
 COSIGN_KEY=${COSIGN_KEY:-}
+GITHUB_ORG=${GITHUB_ORG:-istio}
+DOCKER_HUB=${DOCKER_HUB:-gcr.io/istio-testing}
+
+if [[ -n ${ISTIO_ENVOY_BASE_URL:-} ]]; then
+  PROXY_OVERRIDE="proxyOverride: ${ISTIO_ENVOY_BASE_URL}"
+fi
+
+VERSION="1.14.0-releasebuilder.$(git rev-parse --short HEAD)"
```
ericvn, line 40: I suspect this should also match the version associated with the branch in question, 1.17 for the main branch, 1.16 for this branch. One of the values that isn't getting updated, although other values here seem to have been updated. Have to do more research, but still a number of new 1.14.0-release-builders images created like here: https://console.cloud.google.com/gcr/images/istio-testing/global/istioctl

jjtischer: - release-1.17 branch does not exist for release-builder yet
- the current version in 1.16 is set to 1.14. It seems to me that this test is out of date and is similar enough to the build.sh script that it should be removed and functionality combined with build.sh. 
https://github.com/istio/release-builder/blob/release-1.16/test/publish.sh#L33
-- updated VERSION to be 1.16

ericvn: The release-1.17 branch is currently master (will be cut in January). The process is for changes to be made in master and then cherry-picked. I suspect we need this change in master as well, although it will be a manual cherry-pick with the version.

jtischer-acasi: I just closed this PR and made a new clean on on master. thank you for pointing that out.
https://github.com/istio/release-builder/pull/1319

PR #1332, opened 2023-01-18
Add additional item to the Step=1 branch cut process

ericvn:
I noticed this additional step added in the PR text, but not coded up.


johnma14 (approved):
Thanks Eric.

PR #1383, opened 2023-02-14
Fix helm hub release

howardjohn:


johnma14 (approved):

release/publish.sh
```diff
@@ -40,7 +40,7 @@ VERSION="$(cat "${WD}/trigger-publish")"
 SOURCE_GCS_BUCKET=${SOURCE_GCS_BUCKET:-istio-prerelease/prerelease}
 GCS_BUCKET=${GCS_BUCKET:-istio-release/releases}
 HELM_BUCKET=${HELM_BUCKET:-istio-release/charts}
-HELM_BUCKET_RELEASE=${HELM_BUCKET:-gcr.io/istio-release/charts}
+HELM_HUB_RELEASE=${HELM_HUB_RELEASE:-gcr.io/istio-release/charts}
```
ericvn, line 43: Does line 71 also need a change with this?

howardjohn: yes ty

howardjohn (commented):
fyi this is a regression from https://github.com/istio/release-builder/pull/1254 and partially reverts it, @jtischer-acasi you may need to update your env var name if you are setting this

release/publish.sh
ericvn: Thanks John. Seems like this should be good (adding the gcr.io to what was there).

PR #1406, opened 2023-02-22
Use architecture environment variable for scanner

Kmoneal:
Change architecture during scanning to be configurable.


jacob-delgado:
@howardjohn we don't support arm64 ourselves and are trying to figure out how to specify via parameters the architectures we are building for on our nodes.

howardjohn:
Just for scanner? or everything?

Kmoneal:
I am assuming everywhere else uses `DOCKER_ARCHITECTURE`, this seemed to be the one spot that it was hardcoded. If there is more I am missing, please let me know.

howardjohn (approved):

PR #1409, opened 2023-02-25
Trim brackets if they are included in `ARCH`

Kmoneal:
Trim brackets for DOCKER_ARCHITECTURE if they are present in ARCH

### DIFF
--- a/pkg/build/scanner.go
+++ b/pkg/build/scanner.go
@@ -103,6 +103,9 @@ func Scanner(manifest model.Manifest, githubToken, git, branch string) error {
 	targetArchitecture := os.Getenv("ARCH")
 	if targetArchitecture == "" {
 		targetArchitecture = "linux/amd64,linux/arm64"
+	} else {
+		targetArchitecture = strings.Replace(targetArchitecture, "[", "", 1)
+		targetArchitecture = strings.Replace(targetArchitecture, "]", "", 1)
 	}
 
 	dockerHubs := os.Getenv("HUBS")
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
I'm not against this since it's simple but curious what the use case. is?

Kmoneal:
If `ARCH` is set as a bash array, it will be serialized without bracket which is acceptable for `DOCKER_ARCHITECTURE`. But, if `ARCH` is set as a string with square brackets to be deserialized to an array, similar to the format in [`release/build.sh`](https://github.com/istio/release-builder/blob/fc4fd75a62e0021d62b4823d5ed780c581b5493c/release/build.sh#L42) the square brackets are unexpected characters when the image is pulled later.

This is just not an expected format in building the base images [`export DOCKER_ARCHITECTURES="linux/amd64,linux/arm64"`](https://github.com/istio/istio/blob/c8624737a7c956fa9ff415aa9dec49bdebdff1c8/tools/build-base-images.sh#L48)

Kmoneal:
Please let me know if there is any additional information that would be helpful to explain this change!

howardjohn:
hmm I think it would make more sense to have export DOCKER_ARCHITECTURES="linux/amd64,linux/arm64" be the consistent format. That is how docker, istio/istio, etc all work

Kmoneal:
Would you recommend switching over to `DOCKER_ARCHITECTURES` using its own designated environment variable rather than grabbing `ARCH`? I think that makes sense but seemed a little more of a hurdle if they should be targeting the same architecture

howardjohn:
I didn't neccesarily mean the env name (it may be weird to use "DOCKER_" since not everything is for docker) but the value should all be csv instead of bracketted maybe? so maybe change  release/build.sh?

Kmoneal:
Alright, ARCH now takes in the same format as DOCKER_ARCHITECTURES but transforms it to json format when it is being added to the manifests

howardjohn (approved):

PR #1414, opened 2023-03-02
Trigger build of 1.18.0-alpha.1

howardjohn:
The alpha.0 was buggy


linsun (approved):

ericvn:
We could have rebuilt another .0 since the original .0 wasn't published.

ericvn:
I would think spinning another .0 build and then leaving the .1 out there to be overridden by a successive .1 might be the best way to go.

PR #1415, opened 2023-03-02
fix cosign command for 2.0

howardjohn:
This auto-approvres "By typing 'y', you attest that you grant (or have permission to grant) and agree to have this information stored permanently in transparency logs."

### DIFF
--- a/pkg/publish/docker.go
+++ b/pkg/publish/docker.go
@@ -91,7 +91,7 @@ func Docker(manifest model.Manifest, hub string, tags []string, cosignkey string
 	// able to run 'cosign public-key <key>'.
 	cosignEnabled := false
 	if cosignkey != "" {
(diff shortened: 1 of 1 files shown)


jacob-delgado (approved):

ericvn:
I'm not sure if we want transparency logs for pre-release. We may set upload to them for no? @stewartbutler Thoughts?

If we do want them always, this is a simple fix.

ericvn (approved):

jacob-delgado:
I'm fine with them for alpha and beta builds. That way we can test some of the incoming SLSA requirements as we look at SLSA level 1. cc @kfaseela

howardjohn:
fwiw we always had thenm, it was just implicit before. Signing implies transparency logs (I think)

linsun (approved):

PR #1430, opened 2023-03-16
Generate source code SBOM in 'context' mode

sergii-ssh:
Create SBOM even if the docker output set to context just skip scanning tars.
Add `BillOfMaterialsURI` property to overwrite default location for SBOM


pkg/build/sbom.go
```diff
@@ -26,42 +26,40 @@ import (
 	"istio.io/release-builder/pkg/util"
 )
 
-// Sbom generates Software Bill Of Materials for istio repo in an SPDX readable format.
+const (
+	sourceSpdx    string = "istio-source.spdx"
```
kfaseela, line 30: i was wondering if we should have version appended to the file name... looks like all other artifacts have this. @ericvn thoughts?

sergii-ssh: Good idea, I'll add the version name to the file name.
spdx is create for build/project, i.e combined spdx is generated for all docker tars

howardjohn: We don't version manifest.yaml or a few others fwiw. Its already in a per-version folder (https://gcsweb.istio.io/gcs/istio-release/releases/1.16.0).

howardjohn: but I am fine with it as well

sergii-ssh: updated, ptal

pkg/build/sbom.go
```diff
@@ -26,42 +26,38 @@ import (
...
 
 	// Run bom generator to generate the software bill of materials(SBOM) for istio.
 	log.Infof("Generating Software Bill of Materials for istio release artifacts")
-	if err := util.VerboseCommand("bom", "--log-level", "error", "generate", "--name", "Istio Release "+manifest.Version,
-		"--namespace", releaseSbomNamespace, "--ignore", "licenses,'*.sha256',docker", "--dirs", manifest.OutDir(),
-		"--image-archive", strings.Join(dockerImages, ","), "--output", releaseSbomFile).Run(); err != nil {
-		return fmt.Errorf("couldn't generate sbom for istio release artifacts: %v", err)
+	// For Docker output in 'context' mode we will not produce SBOM.
+	// `bom` can produce bill only for tar and remote images.
+	if manifest.DockerOutput == model.DockerOutputTar {
```
kfaseela, line 51: i am wondering where we are running SBOM in 'context' mode as the PR title says.

pkg/build/sbom.go
```diff
@@ -26,42 +26,38 @@ import (
...
+	releaseSbomNamespace := path.Join(nameSpaceURI, manifest.Version, fmt.Sprintf("istio-source-%s.spdx", manifest.Version))
 
 	// Run bom generator to generate the software bill of materials(SBOM) for istio.
 	log.Infof("Generating Software Bill of Materials for istio release artifacts")
-	if err := util.VerboseCommand("bom", "--log-level", "error", "generate", "--name", "Istio Release "+manifest.Version,
-		"--namespace", releaseSbomNamespace, "--ignore", "licenses,'*.sha256',docker", "--dirs", manifest.OutDir(),
-		"--image-archive", strings.Join(dockerImages, ","), "--output", releaseSbomFile).Run(); err != nil {
-		return fmt.Errorf("couldn't generate sbom for istio release artifacts: %v", err)
+	// For Docker output in 'context' mode we will not produce SBOM.
+	// `bom` can produce bill only for tar and remote images.
```
puerco, line 50: @sergii-ssh  What happens when you build in "context" mode? Happy to learn about the use case if you think a feature is missing. Or is it for local development?

pkg/build/sbom.go
sergii-ssh: Currently we don't call bom at all if context mode is specified. I am enabling it here https://github.com/istio/release-builder/pull/1430/files#diff-e3115b8ca9fdf611ae3dc61dc235e39aeeddbca61036b803ba3c0826f25cc3c9L91

For this snippet. I am skipping sbom generation for docker artifacts since `bom` doesn't support local context/registries but generate one for the source code.

pkg/build/sbom.go
sergii-ssh: hi @puerco In `context` mode images we build are stored in local registry/context, i.e. `gcr.io/istio-release/pilot:1.17.2`. Now if I run `bom generate -o istio-release.spdx --image gcr.io/istio-release/pilot:1.17.2` it will fail because image hasn't been pushed to remote registry yet.

```
FATA generating doc: creating SPDX document: generating SPDX package from image ref gcr.io/istio-release/pilot:1.17.2: while downloading images to archive: fetching remote descriptor: GET https://gcr.io/v2/istio-release/pilot/manifests/1.17.2: MANIFEST_UNKNOWN: Failed to fetch "1.17.2" from request "/v2/istio-release/pilot/manifests/1.17.2". 
```

Workaround is to `docker save` and then run with `--image-archive` but it would be nice to have something like `--image-local` flag to avoid doing so.

ericvn: Can we not point to the images that were created when in context (that's a bad name) mode? They won't been the final location. Or is the SBOM going to end up with inaccurate data? Maybe it would be accurate if we set `sbomOutputURI` to the correct value.

pkg/build/sbom.go
```diff
@@ -26,42 +26,38 @@ import (
 	"istio.io/release-builder/pkg/util"
 )
 
-// Sbom generates Software Bill Of Materials for istio repo in an SPDX readable format.
+const (
+	sbomOutputURI string = "https://storage.googleapis.com/istio-release/releases"
```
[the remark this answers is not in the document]
sergii-ssh: It was hardcoded so I am keeping it not to break anything. This is the default unless manifest.BillOfMaterialsURI is specified. 
I suspect sbom location needs to be specified here as well? https://github.com/istio/release-builder/blob/master/release/build.sh#L38

pkg/build/sbom.go
sergii-ssh: SBOM location is configurable in `manifest.BillOfMaterialsURI` and filenames are stamped with version. Accuracy shouldn't be an issue unless spdx report overwriten on purpose.

pkg/build/sbom.go
kfaseela: OK, so for source code we need to generate SBOM, as docker output mode does not matter in that case. Previously we were fully skipping. @puerco is this correct, that bom cannot support local registries? I remember you mentioning that it also should work.

pkg/build/sbom.go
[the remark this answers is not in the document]
sergii-ssh: Updated build script. Now we'll upload SBOMs to the same bucket as other artifacts.

pkg/build/sbom.go
kfaseela: Could you please clarify how the PR title "[Run SBOM in 'context' mode]" makes sense now? You run it in context mode, but the SBOM is generated only for source code right?

sergii-ssh: Updated the title to highlight that explicitly.

kfaseela:
LGTM

pkg/build/sbom.go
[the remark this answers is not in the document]
jacob-delgado: Can we set this to an environment variable? Vendors and the istio-private org make hard coded URLs difficult to change.

pkg/build/build.go
```diff
@@ -88,9 +88,7 @@ func Build(manifest model.Manifest) error {
 		return fmt.Errorf("failed to package license file: %v", err)
 	}
 
-	if manifest.DockerOutput == model.DockerOutputContext {
-		log.Warnf("Docker output in 'context' mode; will not produce SBOM.")
-	} else if manifest.SkipGenerateBillOfMaterials {
+	if manifest.SkipGenerateBillOfMaterials {
```
ericvn, line 91: We create a new option, SkipGenerateBillOfMaterials, and we output a message if set, but I don't see that we actually act on it. I would have expected some check around the actual BOM generation to not run if this value is set.

pkg/build/sbom.go
ericvn, line 50: We used to generate a message that BOM generation wouldn't happen if we were in context mode and removed that above. If we are in context mode and have not specified to skip the generation, we will not have generated an SBOM, and there would have been no messages in the log that SBOM generation was skipped.

release/build.sh
```diff
@@ -36,6 +36,7 @@ fi
 
 PRERELEASE_DOCKER_HUB=${PRERELEASE_DOCKER_HUB:-gcr.io/istio-prerelease-testing}
 GCS_BUCKET=${GCS_BUCKET:-istio-prerelease/prerelease}
+SBOM_OUTPUT_URI="https://storage.googleapis.com/${GCS_BUCKET}/releases"
```
ericvn, line 39: Do we want this value to be overwrite able as it isn't as written here? If not, then we could include it in the URI below and save a line.

PR #1457, opened 2023-04-13
Using digests rather than image tags for cosign signatures.

stewartbutler:


pkg/publish/docker.go
```diff
@@ -145,7 +145,21 @@ func Docker(manifest model.Manifest, hub string, tags []string, cosignkey string
 			// Sign images *after* push -- cosign only works against real
 			// repositories (not valid against tarballs)
 			if cosignEnabled {
-				if err := util.VerboseCommand("cosign", "sign", "--key", cosignkey, img.NewReference(arch), "-y").Run(); err != nil {
+				imgRef, err := name.ParseReference(img.NewReference(arch))
+				if err != nil {
+					return fmt.Errorf("failed to parse image reference %v: %v", img.NewReference(arch), err)
+				}
+				newImg, err := remote.Image(imgRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```
howardjohn, line 152: Why do we need to get it from the remote? Aren't we the ones pushing it? or we don't really know since its `docker` pushing I suppose

pkg/publish/docker.go
```diff
@@ -240,12 +255,22 @@ func publishManifest(manifest string, images []string) error {
...
 	if err != nil {
-		return fmt.Errorf("failed to parse %v: %v", manifestRef, err)
+		return "", fmt.Errorf("failed to parse %v: %v", manifestRef, err)
 	}
 	if err := remote.MultiWrite(map[name.Reference]remote.Taggable{manifestRef: index}, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
-		return fmt.Errorf("failed to push %v: %v", manifestRef, err)
+		return "", fmt.Errorf("failed to push %v: %v", manifestRef, err)
 	}
-	return nil
+	newImg, err := remote.Image(manifestRef, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```
howardjohn, line 263: On this case I don't think we should need remote.Image, the SHA will be the same before and after we push it afaik?

pkg/publish/docker.go
```diff
@@ -145,7 +145,21 @@ func Docker(manifest model.Manifest, hub string, tags []string, cosignkey string
...
+				if err != nil {
+					return fmt.Errorf("failed to load %v: %v", imgRef, err)
+				}
+				digest, err := newImg.Digest()
+				if err != nil {
+					return fmt.Errorf("failed to get digest for %v: %v", imgRef, err)
+				}
+				// We need to return the digest of the manifest, not the image. This is because the manifest is what is signed.
+				// This should return something like `gcr.io/istio-testing/pilot@sha256:1234`
+				if err := util.VerboseCommand("cosign", "sign", "--key", cosignkey, imgRef.Context().String()+"@"+digest.String(), "-y", "--recursive").Run(); err != nil {
```
ericvn, line 162: if archs==1, do we need --recursive.

ericvn:
LGTM. I'll let @howardjohn approve and then maybe we can kick of an alpha.1 build and verify and try the publish (unless we have a better way to test). I'm not seeing much movement on the branch cut and a beta.0 so maybe an alpha.1 would help for next week?

howardjohn (approved):
seems reasonable to me

PR #1464, opened 2023-04-25
Fix segfault in release automation

pmerrison:
When running release builder automation with DRY_RUN=false, it would segfault with the following error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc594fb]

goroutine 1 [running]:
istio.io/release-builder/pkg/util.PushCommit({{0xc000114540, 0xc0001145a0, 0xc0001147e0, 0xc000114480, 0xc0001146c0, 0xc000114600, 0xc000114840, 0xc0001148a0, 0xc000114660, 0xc000114780, ...}, ...}, ...)
	/work/pkg/util/git.go:92 +0x75b
istio.io/release-builder/pkg/util.CreatePR({{0xc000114540, 0xc0001145a0, 0xc0001147e0, 0xc000114480, 0xc0001146c0, 0xc000114600, 0xc000114840, 0xc0001148a0, 0xc000114660, 0xc000114780, ...}, ...}, ...)
	/work/pkg/util/git.go:149 +0xd45
istio.io/release-builder/pkg/branch.Branch({{0xc000114540, 0xc0001145a0, 0xc0001147e0, 0xc000114480, 0xc0001146c0, 0xc000114600, 0xc000114840, 0xc0001148a0, 0xc000114660, 0xc000114780, ...}, ...}, ...)
	/work/pkg/branch/branch.go:94 +0xe1a
istio.io/release-builder/pkg/branch.glob..func1(0x1a0c120?, {0xc000114300?, 0x6?, 0x6?})
	/work/pkg/branch/cmd.go:71 +0x4a7
github.com/spf13/cobra.(*Command).execute(0x1a0c120, {0xc000114240, 0x6, 0x6})
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x862
github.com/spf13/cobra.(*Command).ExecuteC(0xc000004c00)
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3bd
github.com/spf13/cobra.(*Command).Execute(0x0?)
	/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992 +0x19
main.main()
	/work/main.go:25 +0x53
```

This was because the git-go user struct no longer contains a 'Name' field, replacing it with 'Login' fixes the issue and lets the automation complete.  I've tested this fix on Mac OS 13.3.1 and Debian 11


release/trigger-branch
```diff
@@ -1,2 +1,2 @@
 VERSION=1.17
-STEP=1
\ No newline at end of file
+STEP=1
```
ericvn, line 2: You likely don't want to actually cause a trigger, although in reality there is no automation to cause the branch to actually run. I'd just remove this change.

ericvn:
My concern with this change is we did a similar change for the release-1.17 branching to get things to work, but needed to revert the change (see https://github.com/istio/release-builder/pull/1344) to get the pipeline to work. I'm not sure if we can do a check on the User and decide to use the login or name.

release/trigger-branch
ericvn: In my testing though, I did update the VERSION to 1.18 so that the appropriate branches were used in the steps.

PR #1494, opened 2023-05-23
Remove istio.io/pkg from the manifest

jacob-delgado:
Not sure if we want to remove this as other repos are using this folder to help build istio. Thoughts?

See https://github.com/istio/release-builder/blob/3e098023e5dc6ad036accec5354bdbc2c70b126e/go.mod#L15


jacob-delgado:
See https://github.com/istio/istio/issues/44991

PR #1505, opened 2023-06-06
[release-1.18][Step1] Trigger 1.18.0 GA build

pmerrison:


hanxiaop (approved):
Approve in advance since it has the DNM label.

PR #1519, opened 2023-06-19
Test proxyversion from binary version

cebernardi:
Contributes to [#44937](https://github.com/istio/istio/issues/44937)

This PR changes `TestProxyVersion` check: the version is not checked from the proxy docker image ISTIO_META_ISTIO_VERSION environment variable anymore, but from the binary version (loading the `tar.gz` as docker image, and running the container with `version --short`


howardjohn (commented):
lgtm! thanks

PR #1527, opened 2023-06-23
testing: move test publish to use local storage

howardjohn:
Currently we publsih to GCS and GCR. This moves to using a fake GCS
backend and local docker registry, allowing this to be run without any
privileges. We should still get 99% of the test coverage which is good
enough.

### DIFF
--- a/test/publish.sh
+++ b/test/publish.sh
@@ -27,7 +27,39 @@ else
   echo "No credential helpers found, push to docker may not function properly"
 fi
 
-DOCKER_HUB=${DOCKER_HUB:-gcr.io/istio-testing}
+function cleanup() {
+  # shellcheck disable=SC2046
+  docker stop $(docker ps -a -q --filter label=istio-release-builder)
+}
+trap cleanup EXIT
+
+# Setup fake GCS and registry
+docker run -d  --rm  \
+  -p "7480:5000" --label istio-release-builder \
+  --name "release-builder-registry" \
+  gcr.io/istio-testing/registry:2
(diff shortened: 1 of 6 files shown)


test/publish.sh
```diff
@@ -27,7 +27,39 @@ else
...
+    -H "content-type: application/json" \
+    -H "accept: application/json" \
+    'http://127.0.0.1:7481/storage/v1/b?alt=json&project=test&projection=full' && break
+   sleep 1
+   echo "Trying again... Try #$counter"
+   counter=$((counter+1))
+done
+
+DOCKER_HUB=${DOCKER_HUB:-"localhost:7480"}
+export GCS_HOST=${GCS_HOST-"http://localhost:7481"}
```
ericvn, line 62:
```suggestion
export GCS_HOST=${GCS_HOST:-"http://localhost:7481"}
```

howardjohn: This was intended to allow users to write `GCS_HOST=""` to not used it. But I guess I later check if `host != ""` not if its not found...

howardjohn: Oh actually this is still needed, your suggestion will overwrite `GCS_HOST=""` to `GCS_HOST="localhost:7481"`. The current one will not, but if GCS_HOST is unset it will set it.

TBH I am not sure we need any configuration here since its the test one...

ericvn (approved):
So setting GCS_HOST="" will do not append the host endpoint. Making my suggestion does not allow the user to do that.

PR #1532, opened 2023-06-26
Allow inline docker configuration

howardjohn:
We are no longer using file mounts for secrets. Allow inline config.


ericvn (approved):

howardjohn:
This is not safe. The `set -x` logs the config. Do not use. We will go back to using file

PR #1538, opened 2023-06-26
Allow GH_TOKEN and GITHUB_TOKEN

howardjohn:
Everywhere else uses GH_TOKEN, I think by allowing both it will make things easier to use a consistent var between everything

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The automated `STEPS`:
 
 The following credentials are needed
 
-* Github token: as environment variable `GITHUB_TOKEN` or `--githubtoken file`.
+* Github token: as environment variable `GITHUB_TOKEN`, `GH_TOKEN`, or `--githubtoken file`.
 * Docker credentials (if publishing to docker) (TODO - how to set these).
 * GCP credentials (if publishing to GCS) (TODO - how to set these).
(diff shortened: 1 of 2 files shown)


jacob-delgado (approved):

README.md
```diff
@@ -91,7 +91,7 @@ The automated `STEPS`:
 
 The following credentials are needed
 
-* Github token: as environment variable `GITHUB_TOKEN` or `--githubtoken file`.
+* Github token: as environment variable `GITHUB_TOKEN`, `GH_TOKEN`, or `--githubtoken file`.
```
ericvn, line 94: I'm a little concerned here in that past experience has these being two different tokens and I'm not sure they should be reused. I know one can authenticate `gh` using a GITHIB_TOKEN, but does it work the other way?

howardjohn:
`gh` CLI only used GH_TOKEN as far as I know. I think we can move
everything to GH_TOKEN

ericvn:
gh can use either by doing the auth login with at GITHUB_TOKEN. A quick look and I wonder if we still use the `gh` client.

I know we've run into some issues with tokens in the past, mainly during branch cut, so will want to make sure things work a little before we do the next branch cut.

PR #1539, opened 2023-06-27
Cleanup base image building to use explicit script

howardjohn:
its a bit confusing to have it joined with the current build.sh

`build.sh` can be cleaned up once jobs move to use this method

### DIFF
--- a/release/build-base-images.sh
+++ b/release/build-base-images.sh
@@ -48,6 +48,7 @@ WORK_DIR="$(mktemp -d)/build"
 mkdir -p "${WORK_DIR}"
 
 MANIFEST=$(cat <<EOF
+version: "${VERSION}"
 directory: "${WORK_DIR}"
 dependencies:
   istio:
--- a/release/build-base-images.sh
+++ b/release/build-base-images.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


release/build-base-images.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+# No else needed - the prow entrypoint already runs configure-docker for standard cases
+
+GITHUB_ORG=${GITHUB_ORG:-istio}
+
+WORK_DIR="$(mktemp -d)/build"
+mkdir -p "${WORK_DIR}"
+
+MANIFEST=$(cat <<EOF
+directory: "${WORK_DIR}"
```
ericvn, line 51: I'm noting we have a few items removed from the manifest like version and architectures.

ericvn: It looks like we have defaults for architectures and docker hub, but not seeing a valid default for tag.

howardjohn: Yep, intentional and I tested this locally. These are not needed; keeping them out makes it clear what is used. Even architectures was not used (maybe it should, though).

howardjohn: Good call on tag...

howardjohn: Added back version. Thanks!!

ericvn: Do we want to require a VERSION to be passed or should we default to the value in the trigger-build file?

howardjohn: We want it passed, we set VERSION to the branch name. The one from trigger-build is some random version that has no meaning to base images

ericvn (approved):

PR #1633, opened 2023-09-12
TagRepo(): Explicitly do not sign

jwendell:
Some users have git sign enabled by default, but we do not want this to happen in this job, or within the container shell.


jwendell:
it does: https://git-scm.com/docs/git-tag

ericvn (approved):

PR #1655, opened 2023-10-10
Republish 1.19.2

pmerrison:


hanxiaop:
Looks like we need https://github.com/istio/release-builder/pull/1663?

PR #1668, opened 2023-10-14
Remove envoy dependency in step 1

hanxiaop:
It seems that the Envoy dependency in Istio is not being maintained. 
Actually my version of Envoy is forked from envoyproxy/envoy, and the base branch is not set to 'master' - it's 'main'. This situation is causing the error when running Automation step 1.

### DIFF
--- a/release/branch.sh
+++ b/release/branch.sh
@@ -60,9 +60,6 @@ ${DEPENDENCIES:-$(cat <<EOD
   enhancements:
     git: https://github.com/${REPO_ORG}/enhancements
     branch: ${BASE_BRANCH}
-  envoy:
-    git: https://github.com/${REPO_ORG}/envoy
-    branch: ${BASE_BRANCH}
   proxy:
     git: https://github.com/${REPO_ORG}/proxy
     branch: ${BASE_BRANCH}


ericvn:
I need to look a little more into what might fail in later steps removing the envoy repo from the manifest. I see in https://github.com/istio/release-builder/pull/716 that the build manifest changed the repo.

ericvn:
Places we might need to ignore envoy, We do have nil checks, so things might just work without further changes.
- https://github.com/istio/release-builder/blob/master/pkg/branch/createBranches.go#L35-L37 - comment says envoy should probably be skipped, but logic doesn't have that.
- https://github.com/istio/release-builder/blob/master/pkg/branch/updateCodeOwners.go#L38-L39

ericvn (approved):

hanxiaop:
> Places we might need to ignore envoy, We do have nil checks, so things might just work without further changes.
> 
> * https://github.com/istio/release-builder/blob/master/pkg/branch/createBranches.go#L35-L37 - comment says envoy should probably be skipped, but logic doesn't have that.
> * https://github.com/istio/release-builder/blob/master/pkg/branch/updateCodeOwners.go#L38-L39

Thanks, I'm going to update if anything is breaking when I perform the rest of Automation steps.

PR #1669, opened 2023-10-17
Fix envoy repo in automation step 2, and revise 1.19 to 1.20

hanxiaop:


pkg/model/model.go
```diff
@@ -77,7 +77,6 @@ type IstioDependencies struct {
 	ClientGo       *Dependency `json:"client-go"`
 	TestInfra      *Dependency `json:"test-infra"`
 	Tools          *Dependency `json:"tools"`
-	Envoy          *Dependency `json:"envoy"`
```
ericvn, line 80: This removes the manifest from the manifest, yaml which contains the Envoy sha today (from the envoyproxy/envoy repo). See https://storage.googleapis.com/istio-prerelease/prerelease/1.19.3/manifest.yaml for an example.

release/trigger-branch
```diff
@@ -1,2 +1,2 @@
-VERSION=1.19
-STEP=1
\ No newline at end of file
+VERSION=1.20
```
ericvn, line 1: Fine with updating this, but one can simply set VERSION on the command line before calling the branch.sh. The hope was that we would have automation like we do for the build, so this was the way to set the version. Unfortunately, the automation never occurred.

pkg/source.go
```diff
@@ -48,13 +48,6 @@ func Sources(manifest model.Manifest) error {
 		}
 	}
 
-	// Clone envoy at the end. It needs proxy repo to determine its SHA.
```
ericvn, line 51: I think we still need to clone Envoy as part of the build/publish, need to investigate that some more.

hanxiaop:
Let's wait until the steps are done to determine what can be cleaned up or fixed.

hanxiaop:
> I think we can close this.

Yes, thanks for the review.

PR #1849, opened 2024-05-21
Improve helm chart.yaml parsing a bit, handle dep updates

bleggett:
Putting a hold on this until https://github.com/istio/istio/pull/51174 goes in.


pkg/build/helm_test.go
```diff
@@ -0,0 +1,107 @@
...
+			model.Manifest{
+				Version: "1.19.13-eks-8df270", // sufficiently oddball version string
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			dir, _ := os.MkdirTemp("", tc.name)
+			defer os.RemoveAll(dir)
```
howardjohn, line 51: nit: t.TempDir does this for you

pkg/build/helm.go
```diff
@@ -46,6 +47,9 @@ var (
 	// Currently tags are set as `gcr.io/istio-testing` or `gcr.io/istio-release`
 	hubs = []string{"gcr.io/istio-testing", "gcr.io/istio-release"}
 
+	// The helm repo alias we use to point to the actual Helm repository.
+	releaseRepoAlias = "@istio-release"
```
howardjohn, line 51: TIL...

howardjohn: In our docs we do `helm repo add istio https://istio-release.storage.googleapis.com/charts` though so `@istio` may be more appropriate?

howardjohn: LGTM just pending decision on ^

bleggett: Yeah I was gonna update the docs either way, and figured `@istio-release` was more descriptive, but I don't have a strong opinion.

pkg/build/helm_test.go
bleggett: I couldn't remember if the base testing lib did or our wrapper did, yeah ok.

howardjohn (approved):

pkg/build/helm.go
bleggett: Actually, we don't even need it, we can keep the `file` refs, it's actually simpler, since `helm package` will pack the subcharts into the tgz anyway.

howardjohn (approved):

PR #1871, opened 2024-06-12
Fix ztunnel Helm `globals` special-casing

bleggett:
This goes with https://github.com/istio/istio/pull/51517

Fixes the `release-builder` CI failure there.


bleggett:
This will fail until https://github.com/istio/istio/pull/51517 is in, and https://github.com/istio/istio/pull/51517 will fail until this is in - leaving something of a chicken-egg problem. Suggest merging this first.

PR #1878, opened 2024-06-26
[release-1.22] trigger 1.22.2 build 

thedebugger:


release/trigger-build
```diff
@@ -1 +1 @@
-1.22.0
```
thedebugger, line 1: not sure why this wasn't 1.22.1

zirain (approved):

release/trigger-build
zirain: 1.22.1 is a CVE fix, build from private org.

PR #1892, opened 2024-07-08
Publish istiod-remote Helm chart

nshankar13:
Resolves: https://github.com/istio/istio/issues/51600

### DIFF
--- a/pkg/build/helm.go
+++ b/pkg/build/helm.go
@@ -69,6 +69,7 @@ var (
 		"manifests/charts/istio-cni",
 		"manifests/charts/ztunnel",
 		"manifests/charts/istio-control/istio-discovery",
+		"manifests/charts/istiod-remote",
 		"manifests/charts/ambient",
 	}
 )


bleggett:
1.  `istiod-remote` is basically just `istio-control/istio-discovery` - we literally just `cp` the latter to the former in a Makefile.
2. `istiod-remote` could easily be expressed in terms of a unique Helm profile for `istio-discovery`, and we could drop the chart, the delta is pretty small.

This makes the `istio-discovery` chart a bit more complex - which is mostly because that chart is a mess of obscure + legacy values interacting in non-obvious ways, so adding anything to it makes it worse ATM.

That's a reasonable reason to maintain it as a separate chart, but because we are assembling it via Makefile and `cp` from the regular chart, it _isn't_ actually a fully separate chart, and is mostly just as prone to breakage as if it were actually combined.

All that is to say that we should either merge `istiod-remote` into the regular `istiod` chart, which it is already nearly identical to, or actually maintain it separately - the status quo is sort of the worst of both worlds.

I originally was in favor of merging, but that's liable to be messy without doing surgery on the regular `istiod` chart - so I think I am OK with publishing this as separate.

It will force us to properly decouple the remote chart in the long term, I suspect, which isn't a bad thing.

bleggett (approved):

nshankar13:
Thanks @bleggett. In terms of supportability, are we good in terms of having the remote multi-cluster scenario covered in E2Es, and is it safe to assume that the Helm experience aligns with the remote operator profile? @howardjohn

nshankar13:
And touched on this in the aforementioned issue: https://github.com/istio/istio/issues/51600. But istiod-remote seems a combination of certain manifests from the base chart plus a few from the istiod chart - so both base and Istiod would need to be tweaked to accommodate the remote scenario.

bleggett:
> And touched on this in the aforementioned issue: [istio/istio#51600](https://github.com/istio/istio/issues/51600). But istiod-remote seems a combination of certain manifests from the base chart plus a few from the istiod chart - so both base and Istiod would need to be tweaked to accommodate the remote scenario.

Right - which is sort of a smell, we should probably do that anyway, but I get the argument that that's out of scope.

keithmattix (approved):
LGTM

dhawton (approved):

PR #1903, opened 2024-07-17
Add random bits in branch name to run step few times

thedebugger:


dhawton:
Why would you want to run the step multiple times? You shouldn't be running it more than once.

hzxuzhonghu:
I met same error without this in 1.24

zirain:
> Why would you want to run the step multiple times? You shouldn't be running it more than once.

this will happen when prev branch wasn't deleted.
BTW, only admins are allowed to delete branch.

release/trigger-branch
```diff
@@ -1,2 +1,2 @@
-VERSION=1.19
-STEP=1
\ No newline at end of file
+VERSION=1.23
+STEP=1
```
hzxuzhonghu, line 2: should not change it now

PR #1940, opened 2024-09-20
Trigger 1.22.5 publish, attempt 2

jaellio:
The last one failed partway through. Attempt number 2.


zirain (approved):

zirain:
We don't need this anymore

PR #1958, opened 2024-10-21
[release-1.22] publish 1.22.6

zirain:


release/trigger-publish
```diff
@@ -1,2 +1 @@
-1.22.5
-
```
hzxuzhonghu, line 2: donot remove the newline

zirain: it's fine, that was there because they published twice last time.

zirain:
waitting for long test result

hzxuzhonghu (approved):

PR #2020, opened 2025-01-22
modify update-go-control-plane steps

kfaseela:
Based on https://github.com/istio/test-infra/pull/5584


kfaseela:
@mikemorris feel free to remove dnm when the dependencies are all sorted out

PR #2024, opened 2025-01-30
Specify skip-gar-tagging in the right format

kfaseela:
This is the right format as per https://github.com/istio/test-infra/blob/9e57a458fc9316be18b7c2a2f4c5064b40f665f2/tools/prowgen/cmd/prowgen/main.go#L47


kfaseela:
cc @mikemorris

howardjohn (approved):

mikemorris:
Thanks for catching this @kfaseela! I was still stuck on the earlier segfault so hadn't actually got to the point where this command would be run yet.

PR #2058, opened 2025-03-17
trigger 1.24.4 publish

dhawton:
DNM, hold until #2057 merges and is built.


mikemorris (approved):
Approving, keep DNM label on until we're ready to publish.

PR #2085, opened 2025-04-17
bump golang.org/x/net

dhawton:
Manual handling of #2084


go.mod
```diff
@@ -1,6 +1,8 @@
 module istio.io/release-builder
 
-go 1.23
+go 1.23.0
```
zirain, line 3: we should bump this to 1.24.

dhawton: I don't touch that line manually, and that should only be bumped if we have a legitimate reason to mandate a minimum Go version of 1.24.

jacob-delgado (approved):

PR #2144, opened 2025-07-25
trigger publish 1.26.3

dhawton:
DNM pending test results


zirain:
it should be fine now.

PR #2315, opened 2026-04-27
trigger 1.30.0-alpha3

Stevenjin8:
Depends on https://github.com/istio/release-builder/pull/2316


PetrMc:
Thanks @Stevenjin8  We are building beta today (April 27) Do we need another alpha first for any reason?

Stevenjin8:
> Thanks @Stevenjin8 We are building beta today (April 27) Do we need another alpha first for any reason?

I don't think so.

PR #2357, opened 2026-07-09
Bump Go minimum version to 1.26.0 and update dependencies

fjglira:
- Bump the minimum version of Go required to compile this project from 1.25.0 to 1.26.0
- github.com/go-jose/go-jose/v4 4.1.3 to 4.1.4
- OpenTelemetry, Helm, and other transitive modules updated


go.mod
```diff
@@ -1,19 +1,19 @@
 module istio.io/release-builder
 
-go 1.25.0
```
dhawton, line 3: This isn't "bumping Go", we should only change this if a dependency requires it.. as this is "bumping the *minimum version of Go required to compile this project*"

fjglira: You are right, the wording used was wrong. I updated the text as your suggestion

fjglira: And I updated because to be able to update Helm to the latest version, I get `helm.sh/helm/v4@v4.2.2 requires go >= 1.26.0`

dhawton: Just making sure we aren't doing it arbitrarily.

dhawton (approved):

PR #2367, opened 2026-07-23
Revert default registry in helm charts to docker

Stevenjin8:
I think there is consensus that the cloudflare registry "redirector" is not working out and we should just push to docker. In the future, we should also push to other registries. Will make a blog post about this.


Stevenjin8:
cc @keithmattix

dhawton (approved):

